PR #1, opened 2026-02-06
LSEG PermID scraper

nikki-t:
## Overview

**Related Issue**: [#105 - LSEG PermID Scraper Pipeline](https://github.com/orgs/uchicago-dsi/projects/27?pane=issue&itemId=152520083&issue=uchicago-dsi%7Ccore-facility%7C105)

This PR implements a complete end-to-end pipeline for extracting company information from SEC shareholder data (CIK). The pipeline processes parquet files containing shareholder information, enriches the data with company details from LSEG PermID and GeoNames APIs, and prepares the results for storage in databases or cloud storage.

### Issues Addressed

1. Extract CIK (Central Index Key) identifiers from SEC shareholder tracker parquet files
2. Map CIKs to LSEG PermID identifiers via API
3. Retrieve detailed company information (HQ address, incorporation details, URLs, etc.)
4. Enrich location data using GeoNames API
5. Handle batch processing with progress tracking and resumability
6. Implement threshold-based reprocessing for stale data
7. Provide orchestrator for automated execution
8. Containerize the entire pipeline with Docker
9. Schedule automated pipeline runs with Ofelia
10. Implement comprehensive logging and monitoring

### Pipeline Architecture

Implemented a 4-stage pipeline with modular, reusable components:

**Stage 1: Extract CIKs** → **Stage 2: Query PermIDs** → **Stage 3: Query Company Info** → **Stage 4: Save Results**

### Docker-Compose Containerization

- Orchestrator service with configurable parameters
- Scheduler service for automated execution
- Autoheal service for health monitoring
- Shared network and volumes
- Environment-based configuration

### Data Deduplication

**CIK-Level Deduplication**:
- Removes duplicate CIK entries while preserving unique CIKs
- Handles multiple CIKs per investor correctly
- Maintains data integrity across stages

**Investor-Level Deduplication**:
- Prevents duplicate API calls for same investor
- Tracks all CIKs associated with each investor
- Preserves original investor names

## Unit Testing

- **Total Tests**: 50+
- **Coverage**: >85% code coverage
- **Execution Time**: <30 seconds for full suite
- **Mock Usage**: All external APIs mocked

## Integration Testing

- Manual tests:
    - [x] Parquet with 10 investors, processed in batches of 5 complete successfully within two executions of the orchestrator and correctly identifies no new data to process on a third execution. (Testing entire pipeline to make sure it will process the entire dataset)
    - [X] Processes a new parquet file with new CIKs (and previous CIKs) and confirmed execution on only new CIKs. (Testing pipeline execution on new data)
    - [X] Review execution when the stage 2 PermID query fails, confirmed resumes and is successful when executed again. (Testing batch processing and error recovery)
    - [X] Review execution when the stage 3 company info query fails, confirmed resumes and is successful when executed again. (Testing batch processing and error recovery)

- Testing at scale: See https://github.com/dsi-clinic/idi-company-info/pull/1#issuecomment-3892524069


nikki-t:
**Scale Test Summary**

- Universe size: 7,962 investors (7,964 CIKs). Initial CIK pass runs in about ~1s.

**PermID Retrieval (4 runs: 2026-02-09 to 2026-02-12)**

- Processed in batches of ~2,450 then final 612.
- New PermIDs found by run: 2,283, 2,297, 2,285, 572.
- Total mapped at end: 7,435 investors (7,437 PermIDs).
- Final unmatched: 527 investors (7,962 - 7,435).
- Overall coverage: 93.4% of investors (7,437 / 7,962).
Total runtime: 2h 45m 55s (avg ~41m 29s per run).

**Company Info Retrieval (4 runs: 2026-02-09 to 2026-02-12)**

- Queried from discovered PermIDs each day; final run processed remaining 572.
- End state: 7,437 company records for 7,431 unique investors.
- Difference (6) is due to investors with multiple PermIDs across runs.
- Overall coverage vs full investor universe: 93.3% (7,431 / 7,962).
- Total runtime: 4h 11m 18s (avg ~1h 2m 49s per run).

**Bottom line**

- Pipeline appears stable and near-complete after 4 runs, converging to ~93% coverage.
- Remaining gap is primarily investors without resolved PermIDs, not company-info lookup failures.

docker-compose.yml
alice-duan, line 1: the scheduler/orchestrator pattern is cool! question about how you plan to extend this to many processors--do you envision multiple orchestrator services, each with its own schedule labels to trigger it? or something else?

Makefile
alice-duan, line 1: would it make sense to have these commands be docker compose commands that run in a container instead of on the host system? to more closely match how they would be running in the cloud

Makefile
```diff
@@ -0,0 +1,200 @@
+.PHONY: help install install-dev test clean clean-all stage1 stage2 stage3 pipeline resume-stage2 resume-stage3
+
+# Configuration variables
+# Use .venv Python if it exists, otherwise use system python
+PYTHON := $(shell if [ -f .venv/bin/python ]; then echo .venv/bin/python; else echo python; fi)
```
alice-duan: i think these could be run with `uv run` instead of python/venv

README.md
```diff
@@ -0,0 +1,239 @@
...
+cp /path/to/shareholder_tracker_release.parquet data/watch/
+
+# 3. Start scheduler (runs daily at 2 AM by default)
+docker-compose up -d
+
+# 4. Manual run (executes immediately)
+docker-compose run --rm orchestrator
+```
+
+See [SCHEDULING.md](SCHEDULING.md) for scheduling configuration.
```
alice-duan: i don't see this file

README.md
```diff
@@ -0,0 +1,239 @@
...
+- Threshold: `THRESHOLD_DAYS=30`
+
+**Note:** Changes to `.env` require restarting the scheduler to take effect:
+```bash
+docker compose restart scheduler
+```
+
+For the orchestrator service (manual runs), `.env` changes are picked up automatically on each `docker compose run --rm orchestrator` invocation.
+
+## Alternative: systemd Service
```
alice-duan: i'm unfamiliar with this, what is it and why do we support it?

src/idi_company_info/orchestrator.py
```diff
@@ -0,0 +1,441 @@
...
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from typing import ClassVar, Optional
+
+logging.basicConfig(
+    format='%(asctime)s,%(msecs)d %(module)s:%(lineno)d %(levelname)s %(message)s',
+    datefmt='%Y-%m-%dT%H:%M:%S',
+    level=logging.INFO
+)
```
alice-duan: i think we'd benefit from a logging module so we can call `setup_logging` instead of defining the formats each time

src/idi_company_info/orchestrator.py
```diff
@@ -0,0 +1,441 @@
...
+
+        Args:
+            **kwargs: Arguments to pass to the stage
+
+        Returns:
+            Tuple of (status, error_message)
+        """
+        for attempt in range(1, self.config.retry_count + 1):
+            try:
+                self.logger.info(f"Executing {self.config.name} (attempt {attempt}/{self.config.retry_count})")
```
alice-duan, line 126: thoughts on structured logging instead of formatted strings throughout this project, since there'll be enough going on that we might want to query by attribute?
```
logger.info(
  "Starting exectution",
  extra={
    "name": self.config.name,
    "attempt": attempt
  }
)

File: src/idi_company_info/export_results.py
Comment (alice-duan) [inline src/idi_company_info/export_results.py:1]: super nitty, feel free to disregard, but i was briefly confused between this and `batch_processing.save_results`. might be useful to rename something like `export_result` that distinguishes the purpose of this

File: src/idi_company_info/retrieve_cik.py [hunk 6]
```diff
@@ -0,0 +1,110 @@
...
+
+    # Remove "CIK" prefix from CIK values (e.g., "CIK0001546531" -> "0001546531")
+    subset["investor_cik"] = subset["investor_cik"].str.replace("^CIK", "", regex=True)
+
+    # Remove duplicates AFTER normalization to catch formatting differences
+    subset = subset.drop_duplicates(subset=["investor_name", "investor_cik"])
+    logging.info(f"After normalization and deduplication: {len(subset)} unique investor_name/CIK pairs")
+
+    # Group by investor_name and aggregate CIKs into a list
+    result = subset.groupby("investor_name")["investor_cik"].apply(list).to_dict()
```
alice-duan, line 72: the comments here are super helpful for someone unfamiliar with what the input data looks like 🤩

alice-duan (approved):
lgtm! seems solid, appreciate how well documented this is 🚀  i left a few suggestions and questions for my own understanding. 

broader question about data persistence--so currently intermediate and final results are stored on disk, and `save_result` is a placeholder for writing `company_info.json` to somewhere else. will we ever want to persist intermediate results in the same way, or do they become irrelevant the moment the pipeline run ends? i'm basically wondering if we want to abstract out some sort of storage module that handles all file operations with a universal `load` and `save` function, whether the destination is disk or s3 or elsewhere. then the `save_result` module still does something special to just the final output (e.g. write to postgres). that way if we ever want to change how/where we store things, we can just switch it in the singular storage module. i think you already part way there with the `batch_processing.load_existing_results` and `save` utils, but there are still quite a few `with open...`s scattered throughout

i will leave the weedier parts of the business logic/data to launa 😁 i took a look and it all makes sense to me

alice-duan:
oh also, love that there is a test suite--i'd advocate for running them as a github action that is required before merge, along with a lint check (i have just been using [the clinic one](https://github.com/uchicago-dsi/clinic-template/blob/main/%7B%7B%20cookiecutter.project_slug%20%7D%7D/.github/workflows/main.workflow.yml) on my repos)

nikki-t:
Thanks @alice-dsi, I appreciate the feedback. I will work through each of your comments but for the top level storage suggestion. I think a storage module is a great idea. I put in placeholders for now to give us time to work through the data pipeline and ingest the various processors as I am not entirely sure what that looks like currently.

Can you clarify the with `open...` comment? Are you thinking we would return the data from the file rather than trying to open files individually in the specific processor code?

And also, I plan to define a deployment pipeline that does versioning, lints + formats, tests, scans for vulnerabilities, builds + hosts containers, and deploys infrastructure. The lint, test, and scan will serve as gates to the actual build and deployment.

src/idi_company_info/retrieve_cik.py
```diff
@@ -0,0 +1,110 @@
...
+
+import argparse
+import datetime
+import json
+import logging
+import pathlib
+
+import pandas as pd
+
+logging.getLogger().setLevel(logging.INFO)
```
LaunaG: **Suggestion**

1. Given that logs from different processors will likely be pooled in the same location, it would be helpful to insert context from the larger processing job within the formatting string.  This could be as simple as the job name (e.g., "SCRAPE COMPANY INFO", "DOWNLOAD SEC BULK FILE") but entity identifiers like CIKs could also be inserted when appropriate. To accomplish this pattern, we could use something like `contextvars` or `logging.LoggerAdapter`, but in any case, we'll have to insert the context into the logger somehow. 

2. In the current code, each script creates an instance of the root logger, so configuration steps become repetitive. It would be great to centralize this logic.

3. We are entertaining the idea of streaming logs to CloudWatch.  Ideally, the logger configuration would be dynamic to allow integration with CloudWatch on demand.  For example, using the Python library **[watchtower](https://kislyuk.github.io/watchtower/#)**, we'd have something like:

```python

import watchtower, logging

# Instantiation logic
...

# Handler configuration
logger.addHandler(watchtower.CloudWatchLogHandler())
```

To address these concerns, I think using a factory pattern to instantiate top-level loggers, optionally with CloudWatch configuration, would be helpful. In the past, I've done something like this:

```python
"""Provides loggers for use across the application."""

# Standard library imports
import logging


class LoggerFactory:
    """A simple factory for configuring standard loggers."""

    @staticmethod
    def get(name: str, level: int = logging.CRITICAL) -> logging.Logger:
        """Creates a logger with the given name and level.

        Attaches a stream handler that prints logs in a
        standard format to the console.

        Args:
            name: The logger name.

            level: The initial level. Defaults to 50 ("CRITICAL").

        Returns:
            The logger.
        """
        # Create logger and set level
        logger = logging.getLogger(name)
        logger.setLevel(level)

        # Create console handler and set level
        ch = logging.StreamHandler()
        ch.setLevel(level)

        # Create formatter and add to handler
        format = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
        formatter = logging.Formatter(format)
        ch.setFormatter(formatter)

        # Add handler to logger
        logger.addHandler(ch)

        return logger
```
To provide job context for loggers, you could make loggers additional arguments of your functions (via dependency injection) and then create child loggers if needed. I'm also certain you could keep loggers as global variables instead if you preferred that approach.

docker-compose.yml
nikki-t: I think that after we have ingested the first few processors and understand the input and output of each component in the pipeline then we can revisit the architecture to make a more finalized decision. If we kept things as is and ran on an EC2 then each component could have an orchestrator and a schedule that triggers it. But maybe some components have multiple stages and it would be easier to orchestrate external to the application? I think I want to see how the shareholder and debt tracker processors need to execute and then we can refactor this architecture.

src/idi_company_info/retrieve_cik.py
```diff
@@ -0,0 +1,110 @@
...
+logging.basicConfig(format='%(asctime)s,%(msecs)d %(module)s:%(lineno)d %(levelname)s %(message)s',
+                    datefmt='%Y-%m-%dT%H:%M:%S',
+                    level=logging.INFO)
+
+def get_args():
```
LaunaG, line 21: Nit: Docstrings and type annotations would be nice to have in this script!

Makefile
nikki-t: That's a good point, I am not sure? I can add the docker compose command to run the pipeline. I think I was initially thinking of the Makefile as assisting with running locally and for local development (outside of containerization).

Makefile
nikki-t: Updated to run with `uv`.

README.md
nikki-t: I removed that bit, I don't want to include too much documentation at this point.

README.md
nikki-t: I wanted to find a way to monitor the docker compose process on the EC2 instance. This sets up the orchestrator docker compose service as a [Systemd service](https://en.wikipedia.org/wiki/Systemd) which will help ensure it starts at launch, stays running, provides event logging, and a little bit of monitoring. I updated the readme so that it details this option for docker compose since the orchestrator service gets scheduled and is not running all the time. This is more for future consideration.

src/idi_company_info/orchestrator.py
nikki-t: I am going to leave things as is for now and see how we want to track metrics and logs once we have more of the pipeline complete. Maybe we can include this in the discussion around finalizing the architecture. I think it would be helpful to know how the end user will parse the logs and where they will be located.

src/idi_company_info/orchestrator.py
nikki-t: This has been implemented.

src/idi_company_info/retrieve_cik.py
```diff
@@ -0,0 +1,110 @@
...
+        type=pathlib.Path,
+        required=True,
+        help="Path to output JSON file"
+    )
+    args = parser.parse_args()
+    return args
+
+def read_parquet(input_file):
+    logging.info(f"Reading parquet file: {input_file}")
+    df = pd.read_parquet(input_file)
```
LaunaG, line 42: **Suggestion**

To avoid a tight coupling between local file storage and application code, we could use a library like **[smart_open](https://github.com/piskvorky/smart_open)** that abstracts file locations. Integration with S3, GCP, HTTP, local storage, etc. are provided out-of-the-box:

```python
from smart_open import open

# Stream from a local file
with open('local_file.txt', 'r') as local_file:
    print(local_file.read())

# Stream from an S3 object
# smart_open automatically handles S3 authentication using boto3 mechanisms
with open('s3://my-bucket/remote_file.txt', 'w') as s3_file:
    s3_file.write("Hello, world!")

# Read a gzipped file from S3 transparently
with open('s3://my-bucket/example_file.txt.gz', 'r') as s3_file:
    content = s3_file.read()
    print(content) # The content is automatically decompressed
```

_(Note: File paths would become string types to represent either remote or local locations.)_

I agree with Alice that it would be nice to have a dedicated storage module.  In this case, it could be used to configure the storage library's AWS credentials based on environment settings.

src/idi_company_info/retrieve_cik.py
```diff
@@ -0,0 +1,110 @@
...
+    subset["investor_cik"] = subset["investor_cik"].astype(str)
+
+    # Remove "CIK" prefix from CIK values (e.g., "CIK0001546531" -> "0001546531")
+    subset["investor_cik"] = subset["investor_cik"].str.replace("^CIK", "", regex=True)
+
+    # Remove duplicates AFTER normalization to catch formatting differences
+    subset = subset.drop_duplicates(subset=["investor_name", "investor_cik"])
+    logging.info(f"After normalization and deduplication: {len(subset)} unique investor_name/CIK pairs")
+
+    # Group by investor_name and aggregate CIKs into a list
```
LaunaG, line 71: **Data Question**

Should we group by investor CIK instead of name?  I'm wondering if there are companies with the same name that are only distinguished by their CIK.

src/idi_company_info/utils/api_utils.py
```diff
@@ -0,0 +1,33 @@
+"""API utilities for creating sessions and managing API configurations."""
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+# API configuration constants
+REQUEST_TIMEOUT = (10, 30)  # (connect timeout, read timeout)
+RATE_LIMIT_DELAY = 1.0  # 1 request per second
```
LaunaG, line 9: Is this rate limit specific to the LSEG API?  If so, perhaps it should live with the API logic?

I think this module could also be more generalized to something like "web.py" and not be API specific. (E.g., I use similar logic for scraping webpages.)

src/idi_company_info/query_permid.py
```diff
@@ -0,0 +1,429 @@
+#!/usr/bin/env python3
```
LaunaG, line 1: The code could be made more modular by extracting the "data access layer" for the LSEG API into a separate module that this job calls.  This separation of concerns would also allow us to extend use of the API more easily in the future if we needed to (e.g., adding interfaces to additional API endpoints).

src/idi_company_info/query_permid.py
```diff
@@ -0,0 +1,429 @@
...
+        batch_stats: Statistics from batch processing
+    """
+    # Merge with existing results
+    all_results = {**existing_results, **batch_results}
+
+    # Save results
+    save_results(output_file, all_results)
+
+    # Update batch tracking
+    batch_timestamp = datetime.now().strftime("%Y%m%dT%H%M%S")
```
LaunaG, line 364: Perhaps use UTC for consistency?

src/idi_company_info/export_results.py
```diff
@@ -0,0 +1,357 @@
...
+        # execute_batch(cursor, insert_query, company_data)
+        # conn.commit()
+        # cursor.close()
+        # conn.close()
+        # ```
+
+        self.logger.info("PostgreSQL save completed (placeholder)")
+        return True
+
+    def upload_to_s3(
```
LaunaG, line 155: **Question**

I may have misunderstood your intended implementation; were you thinking that some files would be written locally and others saved to S3?

src/idi_company_info/export_results.py
```diff
@@ -0,0 +1,357 @@
...
+        with open(self.input_file) as f:
+            data = json.load(f)
+
+        if not isinstance(data, list):
+            raise ValueError(f"Expected list of company records, got {type(data)}")
+
+        self.logger.info(f"Loaded {len(data)} company records")
+        return data
+
+    def save_to_postgres(
```
LaunaG, line 68: **Suggestion/Personal Preference**

If we end up persisting processor outputs in a database, I'd recommend structuring the application as a Django project, which scales smoothly due to well-established patterns, handles migrations out-of-the-box, and still permits low-level SQL queries when as needed. Each processor could be a Django app, and each app could have multiple management commands kicked off by an orchestrator, as you have now.

Regardless of whether we go with Django or something like SQLAlchemy + Alembic, I think the code interfacing with the database should go in a separate module.

src/idi_company_info/export_results.py
```diff
@@ -0,0 +1,357 @@
...
+        # Example implementation outline:
+        # ```python
+        # import psycopg2
+        # from psycopg2.extras import execute_batch
+        #
+        # conn = psycopg2.connect(connection_string)
+        # cursor = conn.cursor()
+        #
+        # # Create table if not exists
+        # cursor.execute("""
```
LaunaG, line 122: **Suggestion/Personal Preference**

In my opinion, it would be best to separate data definition language (DDL) SQL commands from insert/update/select commands, especially if the processors end up running continuously. I usually run DDL commands on server startup/initialization before the app starts, but I'm curious if there are other reasons to run it during application execution.

src/idi_company_info/export_results.py
```diff
@@ -0,0 +1,357 @@
...
+import pathlib
+from datetime import datetime
+from typing import Any, Optional
+
+from .utils import get_logger
+
+logger = get_logger(__name__)
+
+
+class ResultExporter:
```
LaunaG, line 24: I'm not sure this level of abstraction is what we're looking for.  Could we stub out separate modules to interface with storage and postgres and then just utilize storage for now?  And the orchestrator would make that call?

src/idi_company_info/export_results.py
```diff
@@ -0,0 +1,357 @@
...
+            connection_string: PostgreSQL connection string
+
+        Returns:
+            True if save successful, False otherwise
+        """
+        self.logger.info("=" * 60)
+        self.logger.info("POSTGRES SAVE (PLACEHOLDER)")
+        self.logger.info("=" * 60)
+
+        if self.dry_run:
```
LaunaG, line 95: I think dry runs would be useful for previewing database migrations (which Django does out of the box!) or expensive or irreversible operations. However, when developing locally, could we test the code against local Docker containers for Postgres and RustFS or MinIO (which implement the S3 API)? This might require us to have more than once Docker Compose file.

src/idi_company_info/query_company_info.py
```diff
@@ -0,0 +1,697 @@
...
+    # Validate batch size
+    if batch_size > len(investors_to_process):
+        logger.warning(
+            f"Batch size ({batch_size}) is larger than investors to process "
+            f"({len(investors_to_process)}). Processing all."
+        )
+
+    return permid_data, existing_results, investors_to_process
+
+def query_geonames_location(
```
LaunaG, line 335: It could also be worth separating the data access layer for GeoNames into a separate module.

src/idi_company_info/utils/api_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def create_session() -> requests.Session:
```
LaunaG, line 12: I can see this logic being helpful to multiple processors.  Do you think it would make sense to have a `common` package for shared functionality?

LaunaG (changes requested):
Great job laying a foundation for the project!  I especially appreciated your extensive documentation and test coverage. Most of my comments relate to opportunities for consolidation and code organization; let me know if it'd be helpful to chat about anything.

nikki-t:
Thanks for the feedback @LaunaG - I had initially created this without understanding the data pipeline requirements so I was really focused in on the CIK to PermID bits. Now that I understand we have to support multiple identifier types plus multiple queries to the PermID API endpoints, I am going to refactor and generalize the code a bit more. I have captured this here: https://github.com/uchicago-dsi/core-facility/issues/120

I will note where that ticket applies in each of your comments so we can trace the feedback. 

I am going to merge this pull request and then work on the new scope which will also include the record matching.

src/idi_company_info/retrieve_cik.py
nikki-t: I am going to factor out a logging module which will better handle logging for all processors.

src/idi_company_info/retrieve_cik.py
nikki-t: I will update those with the next PR.

src/idi_company_info/retrieve_cik.py
nikki-t: I will create a storage module that can use something like `smart_open` to load and return file data.

src/idi_company_info/retrieve_cik.py
nikki-t: I am grouping by investor name across the different search operations for consistency. I do keep a list of multiple CIKs associated with the investor name and each CIK gets queried in the PermID API. Each PermID is retained and then searched to return company info. So company info will ultimately be returned for investors with the same name but differentiated by CIK. We may then want to de-duplicate and do some clean up but I need to run some more tests and see how the data gets passed to the web application before making a decisions on that.

src/idi_company_info/query_permid.py
nikki-t: I will create a API module that covers the different LSEG search queries. Then this script can call those based on the type of identifier passed in.

src/idi_company_info/export_results.py
nikki-t: For now I am going to write intermediate data files between stages to the local file system and then upload them to S3 after successful execution which will reduce the number of API calls made to S3 and help manage failures. With each new execution, the script will pull down the previous data to the file system to pick up where it left off in case of any failures. (If we switch to polling this may need to be re-thought but I would rather tackle that as a separate feature).

src/idi_company_info/export_results.py
nikki-t: This was just a placeholder to keep track of the different output options and a reminder of the Postgres commands. I am going to remove this and we can decide how to implement a common database module once we have figured out its role in the data pipeline.

Can you share an example of a Django project structure you think would fit well for the processors?

src/idi_company_info/export_results.py
nikki-t: As mentioned in the previous comment, we might want to create a separate database module when we have a good understanding of its role in the pipeline.

src/idi_company_info/export_results.py
nikki-t: I will update this script to use the storage module when it is complete and remove the Postgres bits.

src/idi_company_info/export_results.py
nikki-t: I bet we could set something up like that. Let's keep track of this for when we maybe tackle a database common module.

src/idi_company_info/query_company_info.py
nikki-t: I will place this as part of the API common module that includes the LSEG queries.

src/idi_company_info/utils/api_utils.py
nikki-t: I will create a common web module that includes this.

alice-duan:
> Can you clarify the with open... comment? Are you thinking we would return the data from the file rather than trying to open files individually in the specific processor code?

oh yup for this i was just thinking of having all instances of `with open...` that just load or dump json to be standardized behind a util that does that for you in a storage module. so that we only have to change things in one place if we ever decide to do something other than read/write from disk

PR #2, opened 2026-03-02
Implement LSEG PermID identifier pipeline with CIK, CUSIP, and ticker support

nikki-t:
# description

Replaces the old stage-based subprocess pipeline with a unified Identifier class hierarchy that handles PermID retrieval and company info lookup in a single orchestrated run.

**Key changes:**

- `IdentifierCik` and `IdentifierCusip` processors supporting entity search and record match (ticker) query modes
- `Buffer` class for fault-tolerant intermediate storage with configurable flush size
- Batching capped by PermID lookup threshold; skips re-querying already-resolved PermIDs
- Orchestrator driven by `--type {cik,cusip,ticker}` with a registry-based identifier factory
- Docker images updated; Makefile and docs refreshed for new structure
- Package reorganized under `idi_company_info.processors`

**Issues:**

- https://github.com/uchicago-dsi/core-facility/issues/115
- https://github.com/uchicago-dsi/core-facility/issues/120

# tests

All 128 unit tests pass.

Small integration tests (manual execution):

| test                                                                                                       | result                                                                                                                                                                                                                                                                                      | Status |
| ---------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------ |
| 1. Run on batches of 5 for a total of 15 records                                                           | Executed 3 times and processed batches of 5; Fourth execution did not locate any new results.                                                                                                                                                                                               | PASS   |
| 2. Run on a batch of PermIDs but leave company info operations out                                         | Script executed once with PermID generation for 5 entities. Then executed in batches of 5. The script detected the 5 already resolved PermIDs and processed them in the final execution. So they are always executed after all processing has complete and no new identifiers are executed. | PASS   |
| 3. Run on a new file of identifiers (does not contain any old identifiers)                                 | New identifiers were processed and existing data was retained. New data was processed in batches 5 in a total of 3 executions.                                                                                                                                                              | PASS   |
| 4. Run on a file of identifiers that contains overlap with 10 old, already processed identifiers and 5 new | Script executed and located the 5 new identifiers and generated company info for them.                                                                                                                                                                                                      | PASS   |
| Simulate an API error and make sure it does not get added to the do not retry list                         | Simulating a 429 for company info responses means the PermIDs are queued as in test 2 and are processed after all new identifiers are queried and company info has been retrieved.                                                                                                          | PASS   |
| Changed the date for 5 company info records to test the threshold for stale entities                       | Script located 5 stale entities and removed from company info records. Then the script queried for PermIDs and new company info and updated the records with new data.                                                                                                                      | PASS   |
| Set the buffer to 5 and processed 15 entities                                                              | Script saved the PermID and company info results files after 5 entities have been processed for both the PermID and Company Info retrieval.                                                                                                                                                 | PASS   |


# review request

When reviewing the PR keep the following in mind:

1. We will want to revisit the location of the `common` modules in case they are helpful beyond the company info processor and the other processors will be located in a different repository.
2. Track what new functionality to add to the storage module.
3. The S3 storage has *not* been implemented yet but will be in #111 when the S3 bucket is created and the functionality can be tested.
4. For now, the CIK identifier class pulls in the same Parquet file and filters on the data in the format expected for the Shareholder tracker. Need to update for the CDT.
5. The company records will need to be de-duplicated at some point as there may be multiple identifiers to a single entity and that entity may have a slightly different name. This can be implemented here or it can be implemented in a separate job that preps data for the web app.
6. Ticker searching is only available for CUSIP as a back up to try if the entity search does not work. There are many entities to search and one search equals one API request which intermittently locates CIK identifiers. Further integration testing is needed to decide on a direction. We could also factor the ticker operations out and use them as a fall back if an identifier is not located via entity search.
7. Any leftover permids not processed for company information will be retried after all new processing is complete. This priority can be changed based on data requirements.
8. This is currently set up with a schedule, we can figure out how to execute (event trigger or polling) to locate and process new identifiers once we integrate the shareholder and debt trackers. We will also want to pay attention to deployment and make sure new container images are taken into account in container execution.
9. All tests were written by AI but reviewed for coverage so it may not make a ton of sense to delve too deeply into the technical weeds but instead focus on any gaps.
10. This has not been tested at scale as it requires some strategizing and this issue has gotten large in scope. Captured testing details here: https://github.com/dsi-clinic/idi-company-info/issues/3 (Could not create a ticket in the 11th Hour Project so going to leave this as a temporary placeholder)


README.md
```diff
@@ -1,176 +1,271 @@
...
+                    │           │     └── RecordMatchRetriever   (cik-match, ticker)
+                    │           └── Identifier.generate_company_info()  [stage 2]
+                    └── common/
+                          ├── api.py       — LSEG + Geonames API clients
+                          ├── batch.py     — stale/unprocessed entity tracking
+                          ├── buffer.py    — write-buffering to JSON files
+                          ├── failures.py  — permanent-failure registry
+                          ├── logs.py      — structured logging
+                          └── storage.py   — JSON load/save helpers
+```
```
alice-duan, line 159: super helpful 🤩

src/idi_company_info/common/failures.py
```diff
@@ -0,0 +1,157 @@
...
+    LOW_MATCH_SCORE = "low_match_score"  # Match score less than threshold
+    NO_COMPANY_INFO = "no_company_info"  # Entity lookup empty/404
+    API_ERROR = "api_error"  # 5xx, timeout, network
+    RATE_LIMIT = "rate_limit"  # 429
+
+
+class FailureClassifier:
+    """Classifies failures as retryable or permanent."""
+
+    DO_NOT_RETRY = frozenset({FailureType.NO_PERMID, FailureType.NO_COMPANY_INFO, FailureType.LOW_MATCH_SCORE})
```
alice-duan, line 27: this makes sense as a baseline! question for my own understanding: i am curious how permanent "permanent" is. do we know whether the data underlying the api is updated on any regular cadence? wondering if any of these missing-data failure cases would ever wanna be retried like, quarterly if that's when this data set updates or something like that.

i know the pipeline saves these in case we want to do that in the future, but just wanna understand what we know about the lseg data

alice-duan (approved):
lgtm! honestly don't have too much feedback--the organization of common utilities makes a lot of sense to me, and the abstractions are clean and extensible.

some higher level questions:
- why are we supporting cik-match? i thougth ciks were somewhat reliably found with entity search
- why the 2450 batch size? i thought it was close to 5000 a day--is it 2450 for permid retrieval + 2450 for entity lookup, with a bit of a buffer? and then will this just process a batch a day on the scheduled daily run until it makes it through the backlog?
- is this correct: when the shareholder tracker and cdt outputs are ready to be integrated, they will still be in the same parquet format; it's just a matter of discovering/fetching them when they're updated?

> The company records will need to be de-duplicated at some point as there may be multiple identifiers to a single entity and that entity may have a slightly different name. This can be implemented here or it can be implemented in a separate job that preps data for the web app.

i think this can be implemented in a separate job that preps data for the web app!

src/idi_company_info/common/failures.py
nikki-t: I am not sure if that is well documented or not. I can certainly do some searching to try to find out. I think we might also learn a lot about failures in general when we execute a larger test.

We can always add in functionality to track these by date and remove them from the list if a certain threshold date has passed. But I would want to scope that as a separate ticket and I am not sure when it will make sense to implement something - maybe after I do some further testing?

nikki-t:
Thanks @alice-dsi!! Here are my thoughts on your questions:

why are we supporting cik-match? i thougth ciks were somewhat reliably found with entity search
- This would reduce our execution time by half (record match requests allow for up to 1K records) for the initial large batch of identifiers if the record match is as accurate as the entity search but I have to do some testing to find out so both are included for now.

why the 2450 batch size? i thought it was close to 5000 a day--is it 2450 for permid retrieval + 2450 for entity lookup, with a bit of a buffer? and then will this just process a batch a day on the scheduled daily run until it makes it through the backlog?
- Each execution queries the permid API twice so we have to split the 5K request limit in half and then I left some room for retries. This can be configured differently.

is this correct: when the shareholder tracker and cdt outputs are ready to be integrated, they will still be in the same parquet format; it's just a matter of discovering/fetching them when they're updated?
- I am not sure what the CDT output format is and we haven't quite explored whether this processor should poll a directory for new files or if there is some sort of "event" that can trigger its execution.

src/idi_company_info/common/failures.py
alice-duan: yup, definitely out of scope for this pr. was just wondering out loud!

alice-duan:
thanks for the answers! seems like some things we won't know till later, and are therefore out of scope for this PR. it's exciting to see this take shape and become something better defined to iterate off of as the requirements are ironed out!!

src/idi_company_info/common/api.py
```diff
@@ -0,0 +1,250 @@
...
+            headers=headers,
+            data=data,
+            timeout=self.REQUEST_TIMEOUT
+        )
+        response.raise_for_status()
+        return response
+
+
+    def _query_with_error_handling(self, url: str, data: str | dict = None, params: dict = None,
+                                   headers: dict = None, method: str = "get") -> dict[str, Any]:
```
LaunaG: You could use a literal type hint for `method` here:

```python
from typing import Literal

def _query_with_error_handling(self, url: str, data: str | dict = None, params: dict = None,
                                   headers: dict = None, method: Literal["get", "post"] = "get") -> dict[str, Any]:
```

src/idi_company_info/common/api.py
```diff
@@ -0,0 +1,250 @@
+"""Provides API utilities for use across the application."""
```
LaunaG, line 1: Great use of abstraction!

src/idi_company_info/common/api.py
```diff
@@ -0,0 +1,250 @@
...
+            allowed_methods=["GET", "POST"]
+        )
+
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        session.mount("http://", adapter)
+        session.mount("https://", adapter)
+
+        return session
+
+    def get(self, url: str, params: dict = None, headers: dict = None) -> requests.Response:
```
LaunaG: [Optional/Suggestion] It might be helpful to accept a kwargs argument for the `get` and `post` methods to wrap the requests library more flexibly. Future extensions of your methods (i.e., needing to support this or that other requests parameter) would be easier.

pyproject.toml
LaunaG, line 21: It might be helpful to standardize our ruff linting rules. [This configuration](https://github.com/uchicago-dsi/clinic-template/blob/main/%7B%7B%20cookiecutter.project_slug%20%7D%7D/pyproject.toml) is currently used for the Data Clinic.

src/idi_company_info/common/failures.py
```diff
@@ -0,0 +1,157 @@
...
+            identifier: The identifier (CUSIP, CIK, etc.).
+            reason: Optional reason for debugging.
+        """
+        key = (entity_name, identifier)
+        self._entries.add(key)
+        if reason:
+            self._reasons[key] = reason
+        self.save()
+
+    def contains(self, entity_name: str, identifier: str) -> bool:
```
LaunaG: [Question] Do you need both the `contains` and `__contains__` methods?

src/idi_company_info/common/storage.py
```diff
@@ -0,0 +1,54 @@
...
+    Can write in append mode for local files, S3 files are always overwritten.
+
+    Args:
+        file_path: The path to the JSON file.
+        data: The JSON data to save to the file as a dictionary or list.
+    """
+    if "s3://" in file_path:
+        with tempfile.NamedTemporaryFile() as tmp:
+            tp = {'writebuffer': tmp}
+            with smart_open.open(file_path, "w", transport_params=tp) as fout:
```
LaunaG, line 50: [Question] Are you expecting to add authentication to the `transport_params` argument at a later point?

src/idi_company_info/processors/IdentifierCik.py
```diff
@@ -0,0 +1,72 @@
+"""Processes CIK identifiers for company information."""
```
LaunaG, line 1: [Nit] I would make "IdentifierCik.py" and "IdentifierCusip.py" have snake-cased file names.

src/idi_company_info/processors/identifier.py
```diff
@@ -0,0 +1,575 @@
...
+            geonames_api=GeonamesApi(api_key=api_credentials.api_key, geonames_user=api_credentials.geonames_user)
+        )
+
+        self.query_type = query_type
+        self.permid_retriever: PermidRetriever = self._create_permid_retriever(query_type, match_score_threshold)  # Strategy pattern
+
+        self.logger = get_logger("Identifier")
+
+    def _init_dirs(self) -> None:
+        """Initialize the directories."""
```
LaunaG, line 76: Should this logic be agnostic as to whether it occurs locally or for S3?

src/idi_company_info/processors/identifier.py
```diff
@@ -0,0 +1,575 @@
...
+    total_company_info_failed: int = 0
+    duplicates_ids_removed: int = 0
+
+
+class QueryType(StrEnum):
+    ENTITY_SEARCH = "entity_search"
+    RECORD_MATCH = "record_match"
+
+
+class Identifier(ABC):
```
LaunaG: [Nit] I'm not sure this represents a `Identifier` so much as a identifier data pipeline.  Perhaps it could be renamed for clarity? (E.g., `IdentifierPipeline`, `IdentifierWorkflow`, etc.?)

src/idi_company_info/processors/identifier.py
```diff
@@ -0,0 +1,575 @@
...
+    def _parse_permid_entities(self, response: dict) -> list[str]:
+        """Parse the PermID entities.
+
+        Args:
+            response: The API response.
+
+        Returns:
+            The PermID entities.
+        """
+        entities = response.get("data", {}).get("result", {}).get("organizations", {}).get("entities", [])
```
LaunaG, line 363: [Nit] This might be simpler as:

```python
try:
    entities = response["data"]["result"]["organizations"]["entities"]
except KeyError:
    entities = []
```

src/idi_company_info/processors/identifier.py
```diff
@@ -0,0 +1,575 @@
+"""Processes identifiers for company information."""
```
LaunaG, line 1: Nice use of the template pattern!

src/idi_company_info/processors/identifier.py
```diff
@@ -0,0 +1,575 @@
...
+from idi_company_info.common.failures import FailureClassifier, FailureRegistry
+from idi_company_info.common.logs import get_logger
+from idi_company_info.common.batch import BatchProcessing
+from idi_company_info.common.buffer import Buffer
+from idi_company_info.common.storage import load_json, save_json
+from idi_company_info.processors.permid_retriever import PermidRetriever, EntitySearchRetriever, RecordMatchRetriever
+
+
+@dataclass
+class FilePaths:
```
LaunaG: [Optional/Suggestion] I can see an argument for moving these data classes into their own file (e.g., "types.py", "models.py") if they are used across the `processors` package.

src/idi_company_info/processors/identifier.py
```diff
@@ -0,0 +1,575 @@
...
+        """Read parquet file and validate required columns exist.
+
+        Args:
+            input_file: The input file to read.
+            required_columns: The required columns to validate.
+
+        Returns:
+            The dataframe with the required columns.
+        """
+        df = pd.read_parquet(input_file)
```
LaunaG, line 122: Should this be agnostic about reading from S3 or local storage as well? You could do something like:

```python
from smart_open import open
import pandas as pd

s3_uri = 's3://your-bucket-name/path/to/your-file.parquet'

# smart_open provides a file-like object
with open(s3_uri, 'rb') as f:
    # pandas can read directly from the file-like object using the 'pyarrow' engine
    df = pd.read_parquet(f, engine='pyarrow')

print(df.head())
```

src/idi_company_info/processors/IdentifierCusip.py
```diff
@@ -0,0 +1,197 @@
...
+                return f"ticker:{ticker}&&mic:{mic}"
+            else:
+                return f"ticker:{ticker}"
+
+        else:
+            # More than 2 parts - unexpected format, skip
+            return ""
+
+    @staticmethod
+    def _is_bond_security(ticker_str: str) -> bool:
```
LaunaG, line 107: [Question] Oh, interesting; I have a few questions about this.  Are bond securities not present in LSEG?  Is that why we're skipping them here?

src/idi_company_info/processors/IdentifierCusip.py
```diff
@@ -0,0 +1,197 @@
...
+        if len(parts) == 1:
+            # US ticker with no suffix - no MIC specified (let API determine)
+            return f"ticker:{parts[0]}"
+
+        elif len(parts) == 2:
+            # Ticker with exchange suffix
+            ticker = parts[0]
+            exchange = parts[1]
+            # Look up MIC code, default to None if unknown (let API determine)
+            mic = IdentifierCusip.EXCHANGE_TO_MIC.get(exchange, "")
```
LaunaG, line 96: [Question] Does the input data need to be enriched further before it is passed to the LSEG PermID pipeline?

src/idi_company_info/processors/orchestrator.py
```diff
@@ -0,0 +1,359 @@
...
+            query_type=spec.query_type,
+            match_score_threshold=config.match_score_threshold,
+        )
+
+
+# ---------------------------------------------------------------------------
+# Orchestrator
+# ---------------------------------------------------------------------------
+
+class PipelineOrchestrator:
```
LaunaG, line 122: Very clean!

src/idi_company_info/processors/orchestrator.py
```diff
@@ -0,0 +1,359 @@
...
+    ApiCredentials,
+    QueryType,
+)
+from idi_company_info.processors.IdentifierCik import IdentifierCik
+from idi_company_info.processors.IdentifierCusip import IdentifierCusip
+
+
+# ---------------------------------------------------------------------------
+# Status
+# ---------------------------------------------------------------------------
```
LaunaG, line 29: [Nit/Suggestion] My instinct would be to organize the factory, enums, and data classes into one or more files outside of `orchestrator.py`. (Example: Why is `QueryType` imported from `identifier.py` but `IdentifierType` defined here? It is not immediately clear.) I think it could make sense to have something like `registry.py`, `factory.py`, `models.py`, and/or `types.py` or `enums.py` for better modularity.

src/idi_company_info/processors/orchestrator.py
```diff
@@ -0,0 +1,359 @@
...
+    buffer_size: int = 500
+    threshold_days: int | None = None
+    match_score_threshold: int = 1
+
+
+# ---------------------------------------------------------------------------
+# Factory
+# ---------------------------------------------------------------------------
+
+class IdentifierFactory:
```
LaunaG, line 62: Also very clean!

src/idi_company_info/processors/orchestrator.py
```diff
@@ -0,0 +1,359 @@
...
+            return False
+
+        elapsed = datetime.now() - start_time
+        self._log_banner(f"Pipeline completed successfully in {elapsed}")
+        return True
+
+
+# ---------------------------------------------------------------------------
+# CLI
+# ---------------------------------------------------------------------------
```
LaunaG, line 184: [Suggestion] I might separate the CLI code section into `main.py`?

src/idi_company_info/processors/permid_retriever.py
```diff
@@ -0,0 +1,396 @@
+"""Retrieve PermID for company information."""
```
LaunaG, line 1: [Suggestion] It might be helpful to clarify where the work of the strategy ends and that of its caller, the `Identifier` class, begins, in the documentation. It seems that both interact with the `Buffer` class, parse API response objects, and parse the raw data from those objects.  I flipped back and forth between the two files (this and `identifier.py`)  to trace the logic but found the process a bit challenging for a quick scan of the code base.

docker-compose.yml
```diff
@@ -1,5 +1,8 @@
 services:
```
LaunaG, line 1: Not sure if you've already implemented this in the next PR, but you can mock S3 using a Docker service with a compatible API.  I used [RustFS](https://docs.rustfs.com/installation/docker/) recently, and it worked pretty well for testing authorization. If you'd like a reference, I used it for [NBFJA recently](https://github.com/dsi-clinic/nbfja-policy-scorecard/blob/dev/compose.yaml). (Warning: This works but has not been cleaned or reviewed yet haha.)

LaunaG (commented):
Great work!  I appreciated your use of design patterns to make the code more modular.  Most of my feedback is minor, but I had a few larger questions/comments:

1. It looks like the code is still hardcoding I/O to the local filesystem in a few locations. Is that intentional? And should it mock S3 at this point if we expect to use it in the pipeline?

2. I noticed some business logic related to filtering out bonds; is that to accommodate the LSEG API or for another reason?

3. I also saw a mapping from exchanges to MIC.  Does the upstream process need to enrich the data further before it is handed off to the processors?

4. It might be worthwhile to standardize our Ruff configuration with that of the  Data Clinic template repo (link below). I noticed a few inconsistencies in docstrings across the modules.

nikki-t:
@LaunaG - Thanks for the feedback! Here are some answers to your initial top-level statements.

It looks like the code is still hardcoding I/O to the local filesystem in a few locations. Is that intentional? And should it mock S3 at this point if we expect to use it in the pipeline?
- I can't quite remember what locations these might be but I am hoping this is all resolved with this PR: https://github.com/dsi-clinic/idi-company-info/pull/4 (where I updated the S3 and CloudWatch functionality)

I noticed some business logic related to filtering out bonds; is that to accommodate the LSEG API or for another reason?
- In some cases, the ticker field includes what looks like bond data. In order to record match the ticker I need just the ticker name and if possible the exchange so the bond data is left out. I am viewing the record match by ticker as a back up plan and hope that by submitting as many ticker data as possible that we will generate quite a bit of company info versus a more specific search by CIK or CUSIP. But I am in the middle of doing some testing to figure that out: https://github.com/dsi-clinic/idi-company-info/issues/3

I also saw a mapping from exchanges to MIC. Does the upstream process need to enrich the data further before it is handed off to the processors?
- What sort of enrichment are you thinking? This is pretty specific to the record match where it tries to glean the exchange and then translate it to MIC which is what the record match field needs. This is only to retrieve a permid and does not impact the company info. And, so far there is only one exchange identified in the shareholder Parquet file so this isn't really that helpful right now.

It might be worthwhile to standardize our Ruff configuration with that of the Data Clinic template repo (link below). I noticed a few inconsistencies in docstrings across the modules.
- I have set up and defined Ruff in the next PR: https://github.com/dsi-clinic/idi-company-info/pull/4
- I will modify the code for that PR and try to standardize based on the template repo

src/idi_company_info/common/api.py
nikki-t: Update the code to use the literal type hint and simplify conditional selecting method to run.

src/idi_company_info/common/api.py
nikki-t: Implemented in the code.

pyproject.toml
nikki-t: Will standardized in the architecture PR.

src/idi_company_info/common/failures.py
nikki-t: Removed the contains function and updated `batch.py`.

src/idi_company_info/common/storage.py
nikki-t: I was going to rely on an EC2 instance profile and/or IAM permissions (if we diverge from the EC2 architecture) so we do not have to manage credentials.

src/idi_company_info/processors/IdentifierCik.py
nikki-t: I had named them with camel case as they are classes. But maybe that convention has changed?

src/idi_company_info/processors/identifier.py
nikki-t: Maybe this is covered by this PR: https://github.com/dsi-clinic/idi-company-info/blob/bce56e05344b66674468c07de7e0c66cce9297ad/src/idi_company_info/processors/identifier.py#L148-L156

src/idi_company_info/processors/identifier.py
nikki-t: I think I am going to leave this as is because I like how it defines the default empty response.

src/idi_company_info/processors/identifier.py
nikki-t: Pandas can open an S3 URI automatically if `s3fs` is installed which I tested in the architecture PR.

src/idi_company_info/processors/IdentifierCusip.py
nikki-t: I haven't dug too deeply into the LSEG API for bond securities but this is to help parse out the ticker (and mic) fields for the record matching.

src/idi_company_info/processors/IdentifierCusip.py
nikki-t: What sort of enrichment are you thinking? This isn't really buying us much right now and I want to finalize things after I do some comparison testing between the different search types so we might want to leave the exchange stuff out.

src/idi_company_info/processors/identifier.py
nikki-t: Updated to `IdentifierPipeline`.

src/idi_company_info/processors/identifier.py
nikki-t: Created a registry and types file to hold both Identifier data classes and enums as well as Orchestrator.

src/idi_company_info/processors/orchestrator.py
nikki-t: Moved data classes and enums to registry or types files.

src/idi_company_info/processors/orchestrator.py
nikki-t: I think I am going to leave it as is for now. I like keeping the arguments with the code that uses them but will factor out if the orchestrator gets more complicated.

src/idi_company_info/processors/permid_retriever.py
nikki-t: I can update the README. For now the Identifier composes a PermID retriever object since there are two different ways to query for permIds (e.g. entity search, record match). I want to simplify this once I decide which method is best for searching by CUSIP.

docker-compose.yml
nikki-t: Oh interesting, I have only ever used `moto` but I haven't had very complicated S3 use cases. I didn't find a need to mock the S3 since it can also read/write from/to the local file system.

src/idi_company_info/processors/IdentifierCik.py
LaunaG: According to PEP 8, class names should use the CapWords/CamelCase convention, but module and package names should be snake-cased:

https://peps.python.org/pep-0008/#package-and-module-names

src/idi_company_info/processors/IdentifierCusip.py
LaunaG: This was minor, but I'd wondered if the Shareholder Tracker data pipeline (or other upstream process) should add a "MIC" column to the Parquet file before passing it to the LSEG processor, so the processor wouldn't have to manage a mapping itself.  But it seems like we should hold off on that decision until testing has been completed.

docker-compose.yml
LaunaG: I think the reason to mock it would just be to test the application's use of the S3 API locally (e.g., with `smart_open`), without having to interface with AWS directly. Your call, though!  I have not used `moto`.

src/idi_company_info/processors/IdentifierCik.py
nikki-t: Ah I see, I will rename them in the next PR.

PR #4, opened 2026-03-11
Define initial AWS architecture and deployment pipeline

nikki-t:
# Deploy IDI Company Info pipeline to AWS with CI/CD, CloudWatch logging, and S3 integration

## description

Implements the AWS deployment architecture for the IDI Company Info pipeline. Introduces a GitHub Actions CI/CD
workflow, Pulumi infrastructure-as-code, S3-backed intermediate storage, and structured CloudWatch
logging.

**For this PR, it will be helpful to focus on the Pulumi files and to take a look at the `ci.yml` GitHub Actions definition file. Are there best practices that you would like to see implemented in either of these? Is there anything missing?**

**Key changes:**

* GitHub Actions CI/CD workflow: tests, scans, and validates code; builds and pushes Docker images to GHCR/ECR on merge and `issue`
  branch pushes; runs Pulumi `preview` on PRs and `up` on merge to `dev`
* Pulumi infrastructure refactored into modular files (`compute.py`, `iam.py`, `ecr.py`,
  `networking.py`, `secrets.py`, `storage.py`, `user_data.py`) with template-driven EC2 user data
* EC2 launch template with cron-scheduled `pull_and_run.sh`; cron schedules configurable via Pulumi
  stack config
    * Removed Docker-compose scheduler as it seemed simpler to manage cron-style processing on the EC2 instance; this can also pull in any new images without having to rebuild the EC2 instance and interrupt the processor.
* AWS Secrets Manager integration: `PERMID_API_KEY` and `GEONAMES_USER` stored as secrets and
  fetched at EC2 boot time
* S3 input support: orchestrator reads parquet files directly from S3 (`s3://`) in addition to
  local paths
* S3 buffered output: intermediate results written to `s3://idi-dev-processor-s3/output/`

**Issues:**

* https://github.com/uchicago-dsi/core-facility/issues/111

## tests

All 136 unit tests pass.

| test area | count |
| --- | --- |
| `test_api.py` | 8 |
| `test_batch.py` | 3 |
| `test_identifier_cik.py` | 18 |
| `test_identifier_cusip.py` | 34 |
| `test_logging.py` | 14 |
| `test_orchestrator_integration.py` | 28 |
| `test_permid_retriever.py` | 17 |
| `test_storage.py` | 9 |
| `test_buffer.py` + others | 11 |

## review notes

1. Cron schedules default to `02:00` (CIK) and `02:30` (CUSIP) daily and can be overridden via `pulumi config set idi:cron_cik "0 3 * * *"`.
2. The CloudWatch log stream format is `{LOG_STREAM_PREFIX}/{instance_id}/{YYYYMMDD_HHMMSS_ffffff}`. The timestamp is frozen at module import time so all loggers in the same container execution share one stream.
3. S3 output path is `s3://idi-dev-processor-s3/output/`. The bucket is provisioned by Pulumi (`storage.py`) with versioning enabled. If the output path is an S3 URI the orchestrator will store all results data (permid and company info) in S3.
4. ECR images are synced from GHCR on each CI run. The `pull_and_run.sh` script on the EC2 host logs into ECR and pulls the latest image before each cron execution, so new images are picked up automatically without needing to re-provision the instance.
5. The processor job for company info is set up to run entity search for CUSIP and CIK to start, we will want to modify after testing to determine which API to use for which data type and take into account CDT data
6. May want to explore different (better?) credential handling for GitHub Actions deployment to the cloud; current uses GitHub Actions secrets with an AWS access key and secret
7. Need to test versioning scheme when this PR is merged into `dev`


.github/workflows/ci.yml
```diff
@@ -0,0 +1,332 @@
...
+
+      - name: Save outputs
+        id: output
+        run: |
+          V="${{ steps.set-version.outputs.version }}"
+          echo "version=$V" >> $GITHUB_OUTPUT
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            echo "deploy_env=prod" >> $GITHUB_OUTPUT
+          else
+            echo "deploy_env=dev" >> $GITHUB_OUTPUT
```
alice-duan: does this mean `dev` and `release` both deploy to the `dev` stack? is there any further separation here or are they effectively overwriting each other?

i am wondering if we need a `release` branch for this size of project--thoughts on feature branch -> dev -> main?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,332 @@
...
+      - release
+      - 'release/**'
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
alice-duan: does this have implications if pulumi is mid-deploy? i'm unfamiliar pulumi, unsure with how it would handle being cancelled with changes partially applied. but i have run into issues with this in terraform before 😩

.github/workflows/deploy.yml
alice-duan, line 1: this is a really comprehensive workflow! seems like it covers everything we may want to do during a deploy 🤩 

i might recommend splitting it into two workflows, one for pre-merge and one for post-merge (e.g. `pr.yml` and `ci.yml`):
- pre-merge would be a required check that blocks merge and runs unit tests, lint, audits, pulumi preview, etc. basically running validation to make sure the code is "green" before it reaches any CI step that interacts with anything beyond the code itself
   - nitty suggestion: in fact, i would define each of these as a separate required PR check for better visibility. 
- post-merge bumps version, does docker stuff, runs pulumi after we know that any merged changes are already good to go, as gated by the pre-merge workflow

basically, i want to avoid the case where something passes code review, then we merge and trigger this workflow, just to discover that the merged commit fails a unit test, while the change has been deployed via pulumi

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,332 @@
...
+    if: |
+      github.ref == 'refs/heads/main' ||
+      github.ref == 'refs/heads/dev' ||
+      startsWith(github.ref, 'refs/heads/release') ||
+      startsWith(github.ref, 'refs/heads/issue-')
+    defaults:
+      run:
+        working-directory: ${{ env.PROJECT_DIR }}
+    steps:
+      - uses: actions/checkout@v4
```
alice-duan, line 253: curious why we push to ghcr and then sync to ecr before reading from ecr? as opposed to pushing straight to ecr, which is the pattern i am more familiar with

pulumi/README.md
```diff
@@ -0,0 +1,389 @@
...
+pulumi config set idi:geonames_user YOUR_GEONAMES_USERNAME
+```
+
+**Important**: If you configure `permid_api_key` and `geonames_user` here, they will be:
+1. Encrypted in Pulumi state
+2. Stored in AWS Secrets Manager
+3. Automatically retrieved by EC2 instances at launch time
+4. No manual configuration needed after instance launch
+
+If you don't set these values, instances will launch with placeholder values and you'll need to manually update the `.env` file via SSH.
```
alice-duan: any reason not to block on this? i.e. enforce that these are set before instances are launched

pulumi/README.md
```diff
@@ -0,0 +1,389 @@
...
+
+### List Stacks
+
+```bash
+pulumi stack ls
+```
+
+## Backend Configuration
+
+By default, Pulumi uses Pulumi Cloud. To use a different backend:
```
alice-duan: making sure i am understanding correctly--this is set up in ci.yml to use an s3 backend right?

pulumi/infra/compute.py
```diff
@@ -0,0 +1,131 @@
...
+
+from . import config, ecr, iam, networking, secrets, storage, user_data
+
+# -----------------------------------------------------------------------------
+# Config
+# -----------------------------------------------------------------------------
+instance_type = config.config.get("instance_type") or "t2.small"
+_key_name = config.config.get("key_name") or "idi-acct-REMOVED"
+# Omit key_name if placeholder (use SSM Session Manager for access)
+key_name = None if _key_name == "idi-acct-REMOVED" else _key_name
```
alice-duan: why do we do this instead of just `_key_name = config.config.get("key_name")` which would be `None` if `"key_name"` isn't set?

pulumi/infra/compute.py
```diff
@@ -0,0 +1,131 @@
...
+        aws.ec2.GetAmiFilterArgs(name="architecture", values=["x86_64"]),
+        aws.ec2.GetAmiFilterArgs(name="virtualization-type", values=["hvm"]),
+    ],
+)
+
+# -----------------------------------------------------------------------------
+# User Data
+# -----------------------------------------------------------------------------
+_use_s3 = config.config.get("use_s3_output")
+use_s3_output = _use_s3 is None or str(_use_s3).lower() in ("true", "1", "yes")
```
alice-duan, line 34: just curious why we accept all these forms of "yes" lol

does pulumi have an established pattern for boolean config values?

pulumi/infra/compute.py
```diff
@@ -0,0 +1,131 @@
...
+    tags=[
+        {
+            "key": "Name",
+            "value": f"{config.name_prefix}-processor-asg",
+            "propagate_at_launch": True,
+        },
+        {"key": "project", "value": config.project_name, "propagate_at_launch": True},
+        {"key": "environment", "value": config.stack_name, "propagate_at_launch": True},
+        {"key": "managed_by", "value": "Pulumi", "propagate_at_launch": True},
+    ],
```
alice-duan, line 121: nit: could we use config.tags here?

pulumi/infra/compute.py
```diff
@@ -0,0 +1,131 @@
...
+    "idi-processor-asg",
+    name=f"{config.name_prefix}-processor-asg",
+    launch_template=aws.autoscaling.GroupLaunchTemplateArgs(
+        id=launch_template.id,
+        version=launch_template.latest_version,
+    ),
+    vpc_zone_identifiers=networking.default_vpc_subnets.ids,
+    min_size=1,
+    max_size=1,
+    desired_capacity=1,
```
alice-duan, line 111: echoing to check my understanding--the autoscaling group here exists to bring the ec2 instance back up if it fails; it is not currently configured to scale horizontally?

pulumi/infra/iam.py
```diff
@@ -0,0 +1,109 @@
...
+                    "Action": [
+                        "logs:CreateLogGroup",
+                        "logs:CreateLogStream",
+                        "logs:DescribeLogStreams",
+                        "logs:PutLogEvents",
+                        "logs:PutRetentionPolicy",
+                    ],
+                    "Resource": [
+                        "arn:aws:logs:*:*:log-group:idi-ftm2j",
+                        "arn:aws:logs:*:*:log-group:idi-ftm2j:*",
```
alice-duan, line 59: are these the right log groups? the other infra is named like "f"{config.name_prefix}-*", i would expect these to be similar too

pulumi/infra/iam.py
```diff
@@ -0,0 +1,109 @@
...
+                        "Resource": "*",
+                    },
+                    {
+                        "Effect": "Allow",
+                        "Action": [
+                            "ecr:BatchGetImage",
+                            "ecr:GetDownloadUrlForLayer",
+                        ],
+                        "Resource": [
+                            f"arn:aws:ecr:{config.aws_region}:{aid}:repository/{ecr_orchestrator_repo}",
```
alice-duan: pulumi novice question--could we do something like import the ecr_repo defined in `ecr.py` and use `ecr_repo.arn` here? again drawing from terraform experience--i'm worried about manually constructing duplicated arns instead of referencing the actual resource created

pulumi/infra/iam.py
```diff
@@ -0,0 +1,109 @@
...
+
+import pulumi
+
+from . import config
+
+# -----------------------------------------------------------------------------
+# EC2 Role
+# -----------------------------------------------------------------------------
+ec2_role = aws.iam.Role(
+    "idi-role-ssm-agent",
```
alice-duan: nit: we should rename this cuz it does a lot more than ssm, maybe something like "{name_prefix}-ec2-role"?

pulumi/infra/networking.py
alice-duan, line 1: we may want an s3 gateway endpoint so s3<>ec2 doesn't have to go over the internet?

pulumi/infra/storage.py
```diff
@@ -0,0 +1,84 @@
...
+
+from . import config, iam
+
+# -----------------------------------------------------------------------------
+# Processor S3 Bucket
+# -----------------------------------------------------------------------------
+processor_bucket = aws.s3.BucketV2(
+    "idi-processor-s3",
+    bucket=f"{config.name_prefix}-processor-s3",
+    force_destroy=True,
```
alice-duan: curious why we want to force destroy! i would imagine, especially for prod, that we don't want to force destroy a bucket--at least ask for confirmation first if it's not empty

pulumi/templates/secret_retrieval_placeholders.sh
alice-duan, line 1: similar to an above comment, but do we want to add placeholders for required secrets and configs? i would imagine we should just fail if they are missing. same comment for all placeholders throughout the files

alice-duan (commented):
nice! seems like a great start for the terminal, thanks for setting it all up 😁 only change i'd request before merge is splitting the github workflow into pre- and post-merge operations for CI. the rest of the architectural comments i'd like to discuss, but we can iterate on it as the project develops, so i don't want to block this chunk of work on resolving them all right now.

1. i want to clarify the branching strategy. am i understanding this merging progression correctly, with each step happening manually for each change? is the expected flow documented somewhere?
`issue-*` (feature branch, unversioned) -> `dev` (`1.0.0a1`) -> `release` (`1.0.0rc1`) -> `main` (`1.0.0`)

2. apologies for the silly question, but just want to double check we have branch protections set up on this repo? 😅  i don't have permissions to check myself

3. re:
> May want to explore different (better?) credential handling for GitHub Actions deployment to the cloud

  in the past eng orgs i have seen [this pattern](https://aws.amazon.com/blogs/security/use-iam-roles-to-connect-github-actions-to-actions-in-aws/). i am ambivalent on whether we want to implement that here--i do think it's worth it if we will have a lot of similar deploy flows, but i don't want to block this PR on that.

4. disclaimer: i am NOT a networking expert--but is it intentional to have the ec2 in a default subnet (which i think is always public)? and the ec2 could land in any of the default vpc subnets, so we have the vpc endpoints for ssm, ssm messages, and ec2 messages in each of those subnets. i'm wondering the tradeoff of pinning the ec2 to a specific subnet and having the vpc endpoints just for that subnet. we save money but lose the redundancy if an availability zone is down, but for a processing pipeline, it could be worth it since failures are recoverable and downtime isn't a major concern

5. have you provisioned a full aws stack and run a minimal processing script? just to make sure all the networking and permissions are set up, logging is coming through as expected, etc

6. to clarify--are dev and prod stacks in separate aws accounts?

nikki-t:
Thanks for the feedback @alice-dsi! Here are my thoughts are your top-level comments:

nice! seems like a great start for the terminal, thanks for setting it all up 😁 only change i'd request before merge is splitting the github workflow into pre- and post-merge operations for CI. the rest of the architectural comments i'd like to discuss, but we can iterate on it as the project develops, so i don't want to block this chunk of work on resolving them all right now.

- Can you explain what you mean by pre- and post-merge operations?

i want to clarify the branching strategy. am i understanding this merging progression correctly, with each step happening manually for each change? is the expected flow documented somewhere?
issue-* (feature branch, unversioned) -> dev (1.0.0a1) -> release (1.0.0rc1) -> main (1.0.0)

- That is correct! I will update the README (which I think needs a little bit of help anyways)

apologies for the silly question, but just want to double check we have branch protections set up on this repo? 😅 i don't have permissions to check myself

- Yes there is a basic ruleset defined

re:

May want to explore different (better?) credential handling for GitHub Actions deployment to the cloud
in the past eng orgs i have seen [this pattern](https://aws.amazon.com/blogs/security/use-iam-roles-to-connect-github-actions-to-actions-in-aws/). i am ambivalent on whether we want to implement that here--i do think it's worth it if we will have a lot of similar deploy flows, but i don't want to block this PR on that.

- I agree this may be an evolving question. I wonder how the team currently handles credentials? I will explore this and implement whatever we land on but I don't think it has to be for this PR.

disclaimer: i am NOT a networking expert--but is it intentional to have the ec2 in a default subnet (which i think is always public)? and the ec2 could land in any of the default vpc subnets, so we have the vpc endpoints for ssm, ssm messages, and ec2 messages in each of those subnets. i'm wondering the tradeoff of pinning the ec2 to a specific subnet and having the vpc endpoints just for that subnet. we save money but lose the redundancy if an availability zone is down, but for a processing pipeline, it could be worth it since failures are recoverable and downtime isn't a major concern

- The intention is for the EC2 instance to live in a private subnet. I checked the default VPC and it does have public subnets enabled. I am going to set up a separate VPC using Pulumi to enforce this. The VPC endpoints are required to use SSM to access the EC2 on a private subnet. Here is the costs according to a quick web search (needs to be verified):

- Interface Endpoints (PrivateLink):
    - Hourly Fee: ~$0.01 per endpoint per Availability Zone (AZ), roughly $7.30/month per AZ.
    - Data Processing: $0.01 per GB for the first 1 PB processed, decreasing for higher volumes ($0.006 per GB for next 4 PB, $0.004 per GB above 5 PB).
     - Multi-AZ Cost: A typical dual-AZ deployment for high availability costs about $9–$15 per month, plus data fees.

- I will only deploy to a single subnet for now since we want to keeps things under $100/month.

have you provisioned a full aws stack and run a minimal processing script? just to make sure all the networking and permissions are set up, logging is coming through as expected, etc

- Yes I have been testing the search methods on the EC2 and have tested logging and S3 access.

to clarify--are dev and prod stacks in separate aws accounts?

- I am not sure how we will set that up. I think I would prefer them to be separate accounts so we don't mix up our databases and other services but we shall see what the team thinks.

.github/workflows/ci.yml
nikki-t: Yes both `dev` and `release` deploy to the `dev` account since we are only working with two accounts (`dev` and `prod`). I like the idea of creating a release candidate that allows us to deploy and run integration tests on the entire pipeline where dev is used for feature/issue development. This pipeline can change though as we see how the development process works out for different developers.

.github/workflows/ci.yml
nikki-t: I think it could; I will update this to better handle concurrent executions.

.github/workflows/deploy.yml
nikki-t: I believe as it is set up now, the workflow will fail on previous jobs (steps) and not attempt to deploy anything via Pulumi.

Maybe you can clarify what you want the pre-merge to do? How would it block merging?

.github/workflows/deploy.yml
nikki-t: I like hosting containers on GHCR and then push them up to ECR. In the past this has been helpful to maintain the open source side of the container while ECR is private.

pulumi/README.md
nikki-t: I can update this README too, this all gets taken care of in the CI/CD pipeline. But you would need to set those if you are trying to deploy manually from your local dev environment.

pulumi/README.md
nikki-t: Yep Pulumi is set up to use an S3 backend.

pulumi/infra/compute.py
nikki-t: That's a hold over from dealing with secrets. I will fix it.

pulumi/infra/compute.py
nikki-t: I was trying to keep it general for any type of "positive" input. Maybe it is less relevant since the CI/CD pipeline handles this. Just thinking about local deployment and user input and trying to keep that as a backup deployment option in case the CI/CD pipeline fails.

pulumi/infra/compute.py
nikki-t: That reminds me I meant to see if Pulumi has a consistent way to auto tag all resources that takes tag like you can do with Terraform. I will double check this.

pulumi/infra/compute.py
nikki-t: That is correct, it only make sure the EC2 instance is always running.

pulumi/infra/iam.py
nikki-t: Yes at it reduces the complexity in the `logs.py` module. Otherwise I have to pass in the project and stack which I felt would get kind of messy add more arguments to the various stages of the pipeline. If we want to name those with project and stack, I can better explore the best way to do so. I thought it might be okay for the log group names.

pulumi/infra/iam.py
nikki-t: I can try that, I was pondering resource dependencies and when the ARN would be available when the IAM role is created but I might be overthinking and maybe there is way to manage this in Pulumi.

pulumi/infra/iam.py
nikki-t: Sure

pulumi/infra/networking.py
nikki-t: I can define that

pulumi/infra/storage.py
nikki-t: This would be if you run `pulumi down` which would only be run manually. We can leave it out for `prod` but it is helpful in letting me tear down ALL infrastructure after I am done testing.

pulumi/templates/secret_retrieval_placeholders.sh
nikki-t: Not quite sure, I understand the feedback. Maybe I missed a comment?

.github/workflows/deploy.yml
alice-duan: i was thinking of having it set as a required status check on the PR, so the PR can't be merged unless these checks pass.

Screenshot 2026-03-13 at 13 32 42

pulumi/templates/secret_retrieval_placeholders.sh
alice-duan: oops, not sure if that comment saved. basically i am wondering why we set placeholders for secrets and configs that are required for the processor to run. i would imagine we want to fail if they are missing, instead of defaulting to these placeholders.

nikki-t:
I did some digging into the VPC and private subnets. In order to allow outbound traffic from private subnets you have to set up a NAT Gateway which costs ~$30/month which would eat into our $100/month budget.

In AWS, the Internet Gateway requires a public IP on the instance to perform outbound NAT. Without one, outbound traffic to the internet is dropped as there is no way for the traffic to be routed.

The cheapest is to leave things as they are with a public subnet and publicly assigned IP for outbound access. The EC2 has a security group with no inbound rules allowing public traffic and it uses SSM access via endpoints so no SSH access is allowed and there are no services listening publicly. The IAM role follows least-privilege so the EC2 does not have access to all AWS services. The EBS volume is encrypted as well.

I will replace the default security group with one that does not allow any traffic between resources and isolates the EC2 from any other resources defined in that group.

At that point it is fairly secure but best practices is to use a NAT Gateway and private subnet(s). We should evaluate it this when we finalize our architecture with other processors.

nikki-t:
VPC endpoint research:

| Endpoint | Type | Monthly cost | Status |
|---|---|---|---|
| S3 | Gateway | $0 | Add now |
| CloudWatch Logs | Interface | ~$7.30 | Defer |
| ECR API | Interface | ~$7.30 | Defer |
| ECR DKR | Interface | ~$7.30 | Defer |
| **Total if all added** | | **~$21.90/month** | |

CloudWatch keeps the logs private and they do not traverse the Internet but this traffic is encrypted so not a huge risk.

ECR endpoints may make it faster to pull images to the EC2 instances but I don't think we will need to do this very often so that cost outweighs the benefit.

pulumi/infra/compute.py
nikki-t: I had already implemented this. But dropping a note here as a reminder, the Autoscaling Group resource expects tags as a list of dictionaries rather than a single dictionary so it requires slightly different handling than the rest of the resources.

pulumi/templates/secret_retrieval_placeholders.sh
nikki-t: Removed the placeholders.

nikki-t:
Updated the pipeline to use an OIDC provider with two roles one for each pipeline.

.github/workflows/deploy.yml
nikki-t: Checks are implemented.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,154 @@
...
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN_CHECKS }}
+          role-session-name: idi-github-ci
+          aws-region: ${{ secrets.AWS_REGION || 'us-east-2' }}
+
+      - name: Pulumi preview (dev stack)
+        run: |
+          cd pulumi
+          pulumi login s3://${{ secrets.PULUMI_STATE_BUCKET }}
+          pulumi stack select dev 2>/dev/null || pulumi stack init dev
+          pulumi config set aws:region ${{ secrets.AWS_REGION || 'us-east-2' }}
```
LaunaG, line 146: I appreciate your use of pulumi's built-in configuration, as opposed to just injecting environment variables into your infra scripts!  Making a note to myself to follow this pattern moving forward.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,154 @@
...
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+          python-version: "3.13"
+
+      - name: Install Pulumi dependencies
+        run: uv sync --group pulumi
+
+      - name: Configure AWS credentials
```
LaunaG, line 134: Nice--agreed that authentication through OIDC is ideal.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,274 @@
...
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            echo "deploy_env=prod" >> $GITHUB_OUTPUT
+          else
+            echo "deploy_env=dev" >> $GITHUB_OUTPUT
+          fi
+          PULUMI_PROJECT=$(grep '^name:' pulumi/Pulumi.yaml | awk '{print $2}')
+          echo "pulumi_project=$PULUMI_PROJECT" >> $GITHUB_OUTPUT
+
+  docker:
+    name: Build & Push to GHCR
```
LaunaG, line 125: [Opinion/Suggestion] Before noticing Alice's comment, I'd also wondered why we were pushing to both GHCR and ECR, as opposed to the latter only.  It might worthwhile to draft a short [Architectural Decision Record (ADR)](https://github.com/adr/madr?tab=readme-ov-file) (e.g., `docs/adr/infra.md`) to document your decision (and significant architectural decisions more generally) and to link out to any supplemental materials living in Google Drive or Box.

pulumi/infra/compute.py
```diff
@@ -0,0 +1,125 @@
+"""Compute: AMI, launch template, Auto Scaling Group."""
```
LaunaG, line 1: Great job dividing your infrastructure definitions into separate modules for readability and extensibility!  I have not prioritized doing so in my own code yet, but you are motivating me to do so! :D

pulumi/infra/compute.py
```diff
@@ -0,0 +1,125 @@
...
+            ebs=aws.ec2.LaunchTemplateBlockDeviceMappingEbsArgs(
+                volume_size=30,
+                volume_type="gp3",
+                delete_on_termination=True,
+                encrypted=True,
+            ),
+        )
+    ],
+    "user_data": user_data_script.apply(
+        lambda s: __import__("base64").b64encode(s.encode()).decode()
```
LaunaG: [Question] Is there a reason you're lazily loading base64 from the standard library here (as opposed to importing it at the top of the file)?

pulumi/infra/user_data.py
```diff
@@ -0,0 +1,92 @@
...
+    """Load docker-compose.yml for EC2: remove build blocks (EC2 pulls from ECR)."""
+    compose = yaml.safe_load(config.COMPOSE_PATH.read_text())
+    services = compose.get("services", {})
+    for svc in services:
+        if "build" in services[svc]:
+            del services[svc]["build"]
+    return yaml.dump(compose, default_flow_style=False, sort_keys=False)
+
+
+def _load_template(name: str, **replacements: str) -> str:
```
LaunaG, line 31: [Nit] Should the type be omitted for the kwargs here?

pulumi/infra/user_data.py
```diff
@@ -0,0 +1,92 @@
...
+    """Parse cron from config: accept cron format (min hour day month wday) or ofelia (sec min hour day month wday)."""
+    if not raw or not raw.strip():
+        return _ofelia_to_cron(default_ofelia)
+    parts = raw.strip().split()
+    if len(parts) >= _CRON_SCHEDULE_PARTS:
+        return _ofelia_to_cron(raw)
+    return raw.strip()
+
+
+def _load_compose_for_ec2() -> str:
```
LaunaG: [Opinion/Suggestion] To avoid directly manipulating YAML, you could also define multiple Compose files (e.g., `compose.dev.yaml` and `compose.prod.yaml` or `compose.base.yaml` and `compose.override.yaml`, etc.) and then in your Docker Compose command, reference both.  Subsequent files override the values of previously-listed ones:

[https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/)

pulumi/templates/user_data.sh.template
```diff
@@ -0,0 +1,84 @@
...
+cat > .env << ENVFILE_EOF
+# IDI Company Information Pipeline Configuration (matches .env.example)
+PERMID_API_KEY=$PERMID_API_KEY
+GEONAMES_USER=$GEONAMES_USER
+
+# ECR image URI (set by Pulumi)
+ORCHESTRATOR_IMAGE={ORCHESTRATOR_IMAGE}
+
+# Input files — S3 URLs (orchestrator reads directly; docker-compose uses /tmp dummy mount by default)
+INPUT_FILE_ORCHESTRATOR_CIK=s3://idi-dev-processor-s3/input/shareholder_tracker_release_20251218.parquet
```
LaunaG: Should these input file environment variables also be injected?

LaunaG (approved):
Fabulous work!  The processors' deployment pipeline is inherently complex given how much we've been tasked to bootstrap on EC2, but you organized the startup scripts and infrastructure deployments extremely well. I had a few minor comments and suggestions, but overall, it looks good!  As always, feel free to reach out if anything is unclear.

.github/workflows/deploy.yml
nikki-t: I can write up a ADR but we can also decide to only create an ECR image if we are never going to use the GHCR too.

pulumi/infra/compute.py
nikki-t: I had asked AI to help me create some of the resources as it was faster than me reviewing all of the Pulumi resource definitions. I must have missed this upon review and I don't think it makes sense to have an import set up like this for the user_data resource.

pulumi/infra/user_data.py
nikki-t: I think it is okay to leave it as a `str` as Python will interpret it as `dict[str, str]` and for the template all keyword arguments should come in as strings for this particular implementation.

.github/workflows/deploy.yml
nikki-t: Added an ADR.

pulumi/infra/compute.py
nikki-t: Removed the import.

pulumi/infra/user_data.py
nikki-t: This has been implemented.

pulumi/templates/user_data.sh.template
nikki-t: This has been set up.

Issue #8, opened 2026-03-24
Remove entity search from the PermID retrieval operations to simplify the pipeline


nikki-t:
# context

It is clear that the record matching performs at the same level (CIK) or better (CUSIP) and allows for up to 1K identifier queries per request. I would like to modify the company info processor to:

Remove the entity search or simplify the relationship between the PermID Retrieval and Identifier classes. I will leave the entity search in place in case we want that option but it will not be implemented in the identifier classes.
- Use record match by default.
- Add CUSIP as a field to the record match CSV to see if that helps retrieve more accurate information; also only pull in data from the shareholder parquet for CUSIP? Need to test to see if adding a CUSIP is actually considered by the API.

# tasks (optional)

# acceptance criteria

- The Identifier pipeline has been simplified to search using record matching for the permid retrieval operations
- Unit tests pass
- An integration test has been run on 2,400 CIKs and CUSIPs

# out of scope (optional)

Issue #7, opened 2026-03-19
Move the common utilities into a separate repository for the current FTM2J processors


nikki-t:
# context

Each processor maintains its own directory of `common` modules. This should be factored out from both the company information and corporate structure processors and made available in a separate repo that can be installed by both processors.

# tasks (optional)

# acceptance criteria

- The `common` modules are moved to a separate repo
- A PyPI package is defined and can be installed
- The company info processor has been updated to use the new package
- The corporate structure processor has been updated to use the new package

# out of scope (optional)

PR #9, opened 2026-03-26
Remove entity search from the Identifier pipeline and simply code to record match

nikki-t:
# PR: Remove Entity Search — Consolidate on Record Match

**Issue:** dsi-clinic/idi-company-info#8 
**Branch:** `issue-8-remove-entity-search` → `dev`
**Date:** 2026-03-26

## overview

Removes the LSEG Entity Search pipeline entirely and consolidates all identifier
PermID search operations to the LSEG Record Match API. The two supported identifier types are
now `cik` and `cusip` only. The former `cik-match` and `ticker` types have been
eliminated, and the pipeline architecture is significantly simpler as a result.

### summary + review request

This PR simplifies the relationship between the Identifier classes and the different search operations
present in the pipeline. It composes with a Retrieval class for both the PermID and Company Info
retrieval. The strategy pattern was removed with the complicated context handling.

It may be helpful to focus on the `identifier.py` module and the `retrieval.py`, `retrieval_permid.py`,
and `retrieval_company_info.py` scripts as the search functionality was extracted into the retrieval
classes.

### Pipeline architecture (`8c7c64b`, `a67037b`)

- Introduced `PermidRetrieval` and `CompInfoRetrieval` as standalone classes that
  `IdentifierPipeline` composes directly, replacing the previous strategy-based
  `EntitySearchRetrieval` / `RecordMatchRetrieval` split with the pass `Identifier` context object
- Removed `QueryType`
- Removed `LsegEntitySearch` from `ApiClients`; `ApiClients` now holds only
  `record_match`, `entity_lookup`, and `geonames_api`
- Removed `IdentifierType.CIK_MATCH` and `IdentifierType.TICKER`; the registry and
  orchestrator now contain exactly two entries (`CIK` → `IdentifierCik`,
  `CUSIP` → `IdentifierCusip`)
- Retained entity search functionality in `api` module

### CUSIP identifier flow (`75bc832`, `bff8540`)

- `IdentifierCusip.load_data()` now returns `{issuer_name: [cusip, ...]}` (plain
  CUSIP strings) so CUSIP is the stable identifier flowing through `BatchProcessing`,
  failure tracking, and output
- `PermidRetrieval._build_records()` uses CUSIP as `LocalID` and looks up the
  formatted ticker string from `std_ticker_map` as the `Standard Identifier`
- Input parquet for CUSIP now requires three columns: `issuer_name`, `security_cusip`,
  `stock_ticker`
- Warns when a single CUSIP maps to multiple tickers in the input; deterministically
  keeps the first occurrence using `drop_duplicates(subset=["security_cusip"])`
- `CompanyInfo` gains a `ticker: str | None` field — populated for CUSIP runs,
  `null` for CIK runs — keeping the output schema consistent across both types

### Additional updates

- Tests (`c8e27ae`)
- Documentation and execution files (`0c12752`)
- Infrastructure (`3038b57`)

### Unit tests

All 135 unit tests pass with no failures.

```
============================= 135 passed in 0.58s ==============================
```

Coverage spans all modified modules: `retrieval_permid`, `identifier_cusip`,
`identifier_cik`, `identifier`, `batch`, `failures`, `orchestrator`, `factory`,
`registry`.

### Local integration testing

End-to-end pipeline tested locally for both identifier types using real parquet
input files and API responses. Both pipelines produced correct output
structure:

- CIK: `company_info_cik.json` records include `identifier_type: "cik"`, correct
  `identifier` (CIK string), `permid_id`, and `ticker: null`
- CUSIP: `company_info_cusip.json` records include `identifier_type: "cusip"`,
  correct `identifier` (CUSIP string), `permid_id`, and `ticker` populated with
  the raw ticker symbol searched

Failure registry correctly excluded previously failed `(issuer_name, cusip)` pairs
from retry. Deduplication, stale-entity pruning, and batch resumability all verified.

### Cloud deployment and production batch testing

Deployed to cloud environment via Pulumi and executed production batches:

- **CIK pipeline:** batch of 2,450 entities processed successfully; PermID retrieval
  and company info lookup completed; results written to S3
- **CUSIP pipeline:** batch of 2,450 entities processed successfully; CUSIP used as
  `LocalID`, ticker as `Standard Identifier` in Record Match requests; raw ticker
  stored in output; results written to S3


.github/workflows/checks.yml
```diff
@@ -91,7 +91,8 @@ jobs:
       - name: Dependency vulnerability scan (pip-audit)
         run: |
           uv pip install pip-audit
-          uv run pip-audit --desc
+          uv run pip-audit --desc \
```
LaunaG, line 94: Adding a comment to explain what this vulnerability is and why it is being ignored could be helpful.

src/idi_company_info/common/buffer.py
LaunaG: [Nit] This comment could be removed if there are no standard library imports.

src/idi_company_info/common/buffer.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, file_path: str, buffer_size: int = 500, mode: str = "dict") -
         else:
             raise ValueError(f"Invalid mode: {mode}")
 
-    def add(self, data: dict[str, list[tuple[str, list[str]]]]) -> None:
+    def add(self, data: dict[str, list[tuple[str, list[str]]]] | list[dict]) -> None:
```
LaunaG: [Opinion][Low Priority] The shape of the `data` parameter is a bit complex.  One or more type aliases defined outside the class and referenced here could help developers less familiar with the pipeline understand what the data represents.

A quick example from the [Python docs](https://docs.python.org/3/library/typing.html) to illustrate what I mean:

```python
type Vector = list[float]

def scale(scalar: float, vector: Vector) -> Vector:
    return [scalar * num for num in vector]
```

Alternatively, you could reference one of your data classes from `types.py` here

src/idi_company_info/common/buffer.py
LaunaG, line 1: [Question] Should this module be moved to `processors`?  It seems PermID-specific, compared to other modules in the `common` package like `logs` and `storage` that seem generic.

src/idi_company_info/processors/types.py
```diff
@@ -22,7 +22,7 @@ class BatchConfig:
 
     batch_size: int = 2450
     buffer_size: int = 500
-    threshold_days: int = 30
+    threshold_days: int | None = None
```
LaunaG: [Nit] Comments expanding on what `threshold_days` and `buffer_size` represent could be helpful for the uninitiated.

compose.yml
LaunaG, line 1: [Nit] Nice use of base and override Docker files! According to the documentation, the preferred file name is now `compose.yml` (to be less verbose).

README.md
```diff
@@ -57,10 +57,17 @@ export GEONAMES_USER='your-username'
 ### Run
 
 ```bash
+# CIK pipeline
 uv run python -m idi_company_info.processors.orchestrator \
```
LaunaG: [Nit][Low Priority] For simplicity, you could configure a name for your pipeline entrypoint in `pyproject.toml` and then refer to it here like `uv run pipeline`, e.g.:

https://docs.astral.sh/uv/concepts/projects/config/#command-line-interfaces

src/idi_company_info/processors/identifier_cik.py
LaunaG: [Nit/Opinion] For clarity, it could be helpful to keep a `Pipeline` suffix for the subclasses. You could also consider naming pipeline classes according to their output data types (e.g., `CompanyPipeline`, `CompanyByCikPipeline`, `CompanyByCusipPipeline`), if that's not too verbose.

LaunaG (approved):
Great job!  I agree that the code has become simpler and more streamlined following the removal of the other retrieval methods.  I only have a few minor comments (opinions and nits). Feel free to adopt what is relevant and ignore the rest.

src/idi_company_info/common/buffer.py
nikki-t: This is used by the corporate structure processor as is so I think it is generic enough to be used by both but I have created issue #7 to capture the work needed and I will evaluate the common logic vs app specific then.

src/idi_company_info/common/buffer.py
nikki-t: Implemented but will definitely think about where this should live in the common shared repo or the processor specific repo.

PR #19, opened 2026-06-02
Refactor data shape of cache result files

### LINKED ISSUE #17: There is a bug in the company info caching of identifier to permid
# context

**caching**

1. Loads existing results (from company info results file)
2. Uses original entity name and identifier to determine if processed before
3. Then during permid retrieval, loads cached permid file and searches entities to process and locates existing permid (organized by entity name - which can have multiple identifiers with multiple permids)
4. **What happens if an old CIK gets a new PermID?**

**cached identifiers**

To check "have I already got the PermID for this row?" you'd need the PermID, but getting the PermID _is_ the API call the cache is meant to avoid. The only thing you have before that call is `(entity_name, identifier)`, so that has to be the cache key. 

**multiplicity**

- **Multiple PermIDs per entity** — line 89 does `setdefault(entity_name, []).extend(permids)` and line 295 appends, so one name/identifier can resolve to a list of PermIDs.
- **Multiple CIKs per PermID** — per the data model, several CIKs collapse to the same PermID (CIKs are stored as an array per PermID).

So PermID is neither available at cache-check time nor unique against the input rows — it's a many-to-many derived value. `(entity_name, identifier)` is the natural key because it's what uniquely identifies an _input row_ before any API resolution happens.

Today staleness (batch.py filter_stale_entities) prunes only result_file; the cached PermID is reused indefinitely. So if LSEG remaps a CIK to a different PermID upstream, the pipeline silently keeps serving the stale PermID — there is no code path that re-resolves it.

The caching operations needs to be modified to remove the stale entity name and identifier from the Permid file cache:

- Refresh only entities that already have a PermID. Leave genuine never-resolved failures in the do-not-retry registry untouched. (No blanket-retryable, no time-gated failure pass.)
- When a re-resolve returns a different PermID, replace the cached value and emit an INFO log (old→new), since that is the rare, interesting event this feature targets.
- Unify with the existing threshold_days staleness cadence — no separate flag/threshold.

# tasks (optional)

# acceptance criteria

- The Permid ID file cache is also interrogated and stale entries are removed after the entity, identifier have been re-resolved so that ephemeral failures do not overwrite good existing data
- Tests have been updated
- The processor has been run manually and the edge case of a new PermID is tested

# out of scope (optional)

nikki-t:
# refactor(cache): normalize cache shapes, join on permid_url, flag CUSIP false matches (closes #17)

## Summary

Restructures the on-disk caches so the pipeline operates on one flat key space and the two
cache files join cleanly on `permid_url`.

> ⚠️ **Breaking cache change.** The `result_file` shape changed. Delete existing on-disk
> `permid_tracking_*.json` and `company_info_*.json` before running this branch (stale
> entries are otherwise ignored/overwritten as they are reprocessed).

**On-disk shape changes emphasizes the search and the results:**

`permid_file` — owns the input `(name, identifier)` → PermID linkage:
```json
{ "1 800 FLOWERS_cusip_68243Q106": {
    "search": {"Name": "1 800 FLOWERS", "LocalID": "cusip_68243Q106", "Standard Identifier": "ticker:FLWS"},
    "result": ["https://permid.org/1-4295912343"] } }
```

`result_file` — pure company info, keyed by `permid_url` (no entity/identifier linkage):
```json
{ "https://permid.org/1-4295912343": {
    "search": {"permid_url": "https://permid.org/1-4295912343"},
    "result": { ...API fields..., "last_processed": "20260601T143027" } } }
```

The input→PermID relationship lives **only** in `permid_file`; the two files join on
`permid_url`. 

### Reviewer notes
- Focus on the updated shapes, it may be helpful to start with the `company_pipleline.run` and see how previous data is loaded and removed from the input data in `batch.get_unprocessed_entities` and how stale entities are removed `batch.filter_stale_entities`.
- Then you can view the permid retrieval in the `retrieval_permid.retrieve` method which builds and batches the record matches and then queries LSEG. You can also see how a batch or permid data is added to the buffer with `buffer.add`.
- Then you can view the company info retrieval in `retrieval_company_info.retrieve` where the previous data is loaded. In preparation for any failures a separate data structure is created to match the failure module interface and pairs entity name with identifier. Then the permid data is batched and any permids already in the results are removed. Then the company info is retrieved and added to the buffer.
- The buffer is flushed at the end of each retrieval operation (and at intervals during execution) which writes the buffer data to disk.

### Cache normalization (the core refactor)
- **`result_file` is now keyed by `permid_url`** and stores only company info + a `search`
  block, dropping any duplicated input identifier information. Each PermID is now fetched once, 
  written once, overwritten on staleness.
- **`company_pipeline`** builds the flat work-list once and threads `permid_data` into the
  "already processed" decision: a row is processed if its PermID is resolved **and** all its
  `permid_url`s are present in `result_file` (a join across the two caches, not a read of
  per-result identifiers).
- **`batch.py`** — staleness prunes `result_file` and `permid_file` **in sync** via a
  `permid_url → cache_key` reverse index, so stale rows fall back to "needs PermID" and
  re-resolve.
- **`retrieval_company_info`** — consumes the flat permid cache, fetches each unique
  `permid_url` once (deduped), and writes pure `permid_url`-keyed entries.
- **`retrieval_permid`** — simplified to consume/produce the flat shape.
- **`buffer.py`** — dictionary merge simplified to a top-level replace for the new shapes
  (the unique-key invariant for company info makes per-key union unnecessary); flush threshold counts the
  right unit per cache.

### Typed contracts
- **`types.py`** — added `PermidEntry` / `ResultEntry` (+ `PermidSearch`, `ResultSearch`,
  `CompanyResult`) `TypedDict`s documenting the on-disk shapes, applied to the producer
  functions. The buffer stays intentionally generic (`Mapping[str, Any]`) so precisely-typed
  producer output can still be used.

### CUSIP collision check (data-quality, zero extra API calls)
- After a CUSIP run, `find_cusip_collisions` aggregates `permid_file` and warns when one
  PermID is reached by CUSIPs from **more than one issuer** (different 6-char CUSIP issuer
  prefix). Detection only — nothing is pruned or re-queried.

Example log:
```
WARNING  Possible false match: 3 CUSIPs across multiple issuers resolved to PermID
         …4295903265 (Abbott Laboratories): 00258Y104 (ABACUS GLOBAL MANAGEMENT INC),
         002824100 (ABBOTT LABORATORIES), 89628W708 (ABACUS FCF REAL ASSETS LEADERS ETF)
WARNING  CUSIP collision summary: 8 PermID(s) reached by multiple CUSIP issuers (review)
```

## Validation
- `uv run pytest` — **139 passing**; `uv run ruff check src/ tests/` clean.
- Cloud integration runs (CIK + CUSIP) inspected: CIK is clean 1:1 (1,401 → 1,401); CUSIP
  collapses 1,261 rows → 269 PermIDs, and the collision check flags 8 multi-issuer suspects
  while correctly leaving 4 legitimate share-class collapses unflagged.


nikki-t:
# results of testing in the cloud

|  | CUSIP | CIK |
| -- | -- | -- |
| Input rows (cache keys) | 1,261 | 1,401 |
| Distinct identifiers | 283 CUSIPs | 1,401 CIKs |
| Distinct names | 1,204 | 1,401 |
| Unique PermIDs (result entries) | 269 | 1,401 |
| Multi-identifier→one-PermID collapses | many | 0 |

-  For CUSIP, genuinely different companies can collapse into one PermID, and it's a Record Match quality problem as they return a match score of 100%.
- **The CIK data is 1:1 — no collapse at all.** 1,401 rows → 1,401 CIKs → 1,401 names → 1,401 PermIDs → 1,401 result entries. Every input row is a distinct CIK with a distinct name resolving to a distinct company.

src/idi_company_info/batch.py
```diff
@@ -1,184 +1,218 @@
...
+            for url in entry["result"]:
+                url_to_keys[url].append(key)  # Reverse index: permid_url -> [permid_cache_key, ...]
+
+        stale_keys = {key for url in stale_urls for key in url_to_keys.get(url, [])}
+
+        # Prune both caches in place.
+        for url in stale_urls:
+            self.result_data.pop(url, None)
+        for key in stale_keys:
+            self.permid_data.pop(key, None)
```
alice-duan: is it correct to unconditionally prune the entire `permid_data` key here? 

for example: if stale URL `https://permid.org/1-abc` points to key `Acme_cik_1234`, which has `result: [https://permid.org/1-abc, https://permid.org/1-xyz]`, then deleting the entire `Acme_cik_1234` would delete the only reference  in `permid_data` to `https://permid.org/1-xyz` (if no other cache key points to it), without pruning its entry from the `result_data`.

would it make more sense to just remove the stale url if there are several:
```
for key in stale_keys:
  if len(self.permid_data["key"]["result"]) == 1:
    self.permid_data.pop(key, None) # remove entire key if the only url is stale
  else:
    self.permid_data["key"]["result"].remove(stale_url) # remove only stale url if there are several urls

File: src/idi_company_info/cache_keys.py [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+        key: The flat permid_file key.
+
+    Returns:
+        Tuple of (entity_name, identifier_type, identifier).
+    """
+    parts = key.rsplit("_", 2)
+    if len(parts) != _EXPECTED_KEY_PARTS:
+        raise ValueError(f"Malformed permid cache key: {key!r}")
+    entity_name, identifier_type, identifier = parts
+    return entity_name, identifier_type, identifier
```
alice-duan: claude points out that this is uncalled--was it meant to be called somewhere or should we delete it?

src/idi_company_info/company_pipeline.py
alice-duan, line 263: i think we may want to consult the failure registry here--if `(Acme, cik_1234)` was written to the failure registry on a previous run due a successful permid lookup but failed company info call, then it is removed from `batch._remove_failed_entities`. but it (correctly) still exists in `permid_data`, so it will be retried here unless we remove entities from the failure cache at this step.

```
if self.failure_registry:
  permid_data = {
    key: entry for key, entry in permid_data.items()
    if (entry["search"]["Name"], entry["search"]["LocalID"]) not in self.failure_registry
  }
```

src/idi_company_info/types.py
```diff
@@ -4,6 +4,60 @@
...
+    """A permid_file entry: the input search block and the resolved PermID URLs."""
+
+    search: PermidSearch
+    result: list[str]
+
+
+class ResultSearch(TypedDict):
+    """The search block of a result_file entry."""
+
+    permid_url: str
```
alice-duan: just curious--why is this necessary? is it mostly to parallel the shape of the permid file with the `"search"` field? it's redundant with the key, but it's not too cumbersome to duplicate it here, so mostly wondering out loud here

src/idi_company_info/batch.py
```diff
@@ -1,184 +1,218 @@
...
+
+        return len(self.result_data)
+
+    def _get_stale_urls(self) -> set[str]:
+        """Return the set of permid_urls whose result is older than threshold_days."""
         if self.threshold_days is None:
-            return set(), []
+            return set()
 
         threshold_date = datetime.now() - timedelta(days=self.threshold_days)
```
alice-duan: very minor bug: since `datetime.now()` matches the system's local time, but we write in `retrieval_company_info._parse_company_info` (line 205) with `tz=UTC`, i think we want to use `datetime.now(tz=UTC)` here to ensure timezone consistency. 

if we want to be maximally explicit, we can use `.isoformat()` in place of `strftime` (and `.fromisoformat()` in place of `strptime`) on line 180 below and in `retrieval_company_info._parse_company_info` (line 205), since `isoformat()` will include the UTC offset `"2026-06-01T21:45:12+00:00"`

alice-duan (approved):
lgtm! the new shapes make so much sense, it seems obvious in hindsight 😅 thanks for working through this change!

if i understand correctly, this PR changes `permid_data` from an internal cache/intermediate step in creating `result_data` to something consumer facing, since consumers now have to read from both to get the same info that was only in `result_data` before. just wanted to call this out and clarify my understanding around that? ~~alternatively, we could treat both `permid_data` and `result_data` as processor intermediates and generate a final output that denormalizes these into one single file, depending on what makes the most sense for the nature of this dataset and how it may be consumed.~~ (nvm, talked myself out of that as i was typing. i think we established earlier today that these will be consumed separately)

high level nit--i wonder if it would be more intuitive to call the outputs something other than `permid_data` and `results_data`, in both the codebase and for output files. i got a bit tripped up on what each of these meant, as someone who hasn't spent as much time as you have in the weeds on this, especially on some types like `ResultSearch`. thoughts on something like `permid_matches` and `permid_entities`?

nikki-t:
> high level nit--i wonder if it would be more intuitive to call the outputs something other than permid_data and results_data, in both the codebase and for output files. i got a bit tripped up on what each of these meant, as someone who hasn't spent as much time as you have in the weeds on this, especially on some types like ResultSearch. thoughts on something like permid_matches and permid_entities?

Thanks for the review @alice-dsi , I appreciate it! This is a good question maybe call it `identifier_data` and `company_data`? I am also wondering if we should key into the "results" part of the cache and call it something like `permid_url_data` and `company_metadata`? I don't think I will change this in this PR as the next development phase covers input sources and saving the final output into parquet files.

src/idi_company_info/cache_keys.py
nikki-t: Thanks for finding that - it was called in a previous iteration but it's no longer needed. It make the `cache_keys module pretty sparse so I am going to place the remaining `permid_cache_key` in the `buffer` module.

src/idi_company_info/batch.py
nikki-t: This was a good catch!

Staleness is measured on company info, not on the permid mapping so I think it makes sense to remove the `permid_data` pruning. If we want to mark permid URLs as stale then we should implement that separately from the company info data pruning. I think it might be rare for the identifier --> permid mapping to change.

So, if I just remove the stale company info then the retrieval of the permid URL is not re-queried and only the company info is retrieved again.

There is filtering along the way that make sure the removed URL is processed while any existing fresh URLs for the same identifier are not.

src/idi_company_info/types.py
nikki-t: I was trying to make it clear what the types were for the different results as I was running into typing errors. I think this is a leftover and doesn't really lend any functionality. I will remove it.

src/idi_company_info/company_pipeline.py
nikki-t: Yes, I have updated the code.

src/idi_company_info/types.py
nikki-t: The result_file output shape is now `{permid_url: {investor_name, permid_id, …, last_processed}}`

PR #22, opened 2026-06-15
Define input sources and aggregate final output to parquet

### LINKED ISSUE #16: Connect the FTM2J terminal processors to the company info processor
# context

The company info processor can take identifiers from the other FTM2J terminal processors and return a PermID. The output storage for each processor needs to be formalized and the company info processor needs to be modified so that it can read in and locate identifiers from the output files.

The company info processor may also need to be modified to output the results in a way that makes sense; this may involve changing the output file names to be named after each processor so it is clear which PermIDs are for what processor. It would be good to consult with the UI to make sure the output of the company info processor is organized in a way that works for the UI ingestion operations.

In addition to the above, it may also need to be modified to execute based on processor type - currently it does so by id type but that may need tweaking as the assumption was to execute on CUSIP for Shareholder Tracker and CIK for CDT but the corporate structure processor also produces CIKs.

# tasks (optional)

- The other processors (Shareholder, CDT, Corporate Structure) are placing output files in a common location (S3).
- The company info processor can read in these files and locate identifiers and query the LSEG API for PermIDs.
- The company info schedule is modified to execute once for each processor type.
- The company info architecture supports the addition of processor types rather than identifier types and there is one pipeline per processor.
- Unit tests pass and linting + formatting pass.
- The processor has been deployed and tested in the cloud.

 
# acceptance criteria

# out of scope (optional)

nikki-t:
# Refactor pipeline around pluggable input sources and aggregate final output to parquet

**Issues:**

* https://github.com/dsi-clinic/idi-company-info/issues/16

## description

Restructures the company-info pipeline around a composable input-source pattern, adds
support for all four input sources (shareholder CIK, shareholder CUSIP, commercial debt,
corporate subsidiaries), updates CDT to read a single parquet file, and introduces a
lock-protected final aggregation step that combines every processor's cache into a single
output parquet.

**Reviewer focus**

- Review the input source classes in `input.py` and evaluate solution for easy addition of sources
- Review `output.py` and locking mechanism - This is not a foolproof method to prevent processors from overwriting each other but helps with any overlap in scheduled company info processor execution and final aggregation of the output data; we should finalize the different processor schedules and decide on any refinements

**Pipeline refactor — compose input source with pipeline:**

* Removed `CompanyByCikPipeline` and `CompanyByCusipPipeline` subclasses; replaced with a
  single `CompanyPipeline` that accepts a pluggable `Input` loader via dependency injection
* Added `input.py` with four `Input` subclasses — `ShareholderInputCik`,
  `ShareholderInputCusip`, `CdtInput`, `SubsidiaryInput` — each responsible for reading
  and normalising its source parquet
* Added `registry.py` (`INPUT_REGISTRY`) mapping each `InputSource` enum value to its
  `Input` class; adding a new source now requires only a registry entry
* `PipelineFactory` selects and instantiates the correct `Input` from the registry and
  composes it into `CompanyPipeline` at build time
* Removed the `failure_file` argument from `OrchestratorConfig`; failure files are derived
  from the output subdirectory alongside the other cache files
* CIK values from the CDT and corporate-subsidiaries inputs are zero-padded to 10 digits
  to match the format expected by the Record Match API

**Final aggregated parquet output:**

* Added `output.py` — `Output.aggregate()` iterates every `InputSource`, joins each
  processor's `permid_url.json` against its `permid_data.json` on the PermID URL (left
  join — one row per identifier→PermID link; unresolved PermIDs kept with null company
  fields), concatenates across all processors, and writes a single `company_info.parquet`
  with a stable column schema
* Added `lock.py` — `FileLock` context manager serialises writes to the final parquet:
  `fcntl.flock` for local paths (covers multiple processes on one host); S3 conditional
  `PutObject` (`If-None-Match: *`) for `s3://` paths (covers multiple hosts); abandoned
  locks older than a configurable TTL are stolen so a crashed process cannot deadlock
  subsequent runs
* Added `paths.py` — `join_path()` extracted from `factory.py` and shared between the
  factory and output modules for consistent local/S3 path joining
* `Output.aggregate()` runs automatically after each successful `pipeline.run()` and is
  also exposed as a standalone `aggregate` CLI entry point which is also an ECS tasks that
  can be run manually - we may want to add as a schedule later
* `OrchestratorConfig` gains `final_output_file` and `skip_final_output`; corresponding
  `--final-output-file` and `--skip-final-output` CLI flags added to the orchestrator
* `boto3` added as an explicit dependency (previously only transitive) to support the
  self-contained S3 lock client

**Pulumi / scheduling:**

* Pulumi stack config refactored into `Pulumi.dev.yaml` and `Pulumi.prod.yaml` with
  per-source `input_file`, `cron`, and `batch_size` entries driven by `idi:input_sources`
* `scheduling.py` updated to schedule one EventBridge rule per input source entry;
  currently active sources: `shareholder_tracker_cik`, `commercial_debt_tracker`,
  `corporate_subsidiaries`
* Removed the GitHub Actions deploy workflow and checks workflow (superseded by Pulumi-driven
  scheduling)

## tests

* `test_input_shareholder_cik.py`, `test_input_shareholder_cusip.py`,
  `test_input_commercial_debt.py`, `test_input_corporate_subsidiaries.py` — unit tests for
  each `Input` subclass covering filtering, deduplication, zero-padding, and column
  validation; CDT tests updated from shard-directory to single-file fixtures
* `test_company_pipeline.py` — tests for the composed `CompanyPipeline` base class
* `test_orchestrator_integration.py` — end-to-end tests for all four input sources;
  CDT integration test updated from `cik_shard=0000/part-0000.parquet` to a single
  parquet file
* `test_output.py` — tests for `Output.aggregate()` covering processor mixing, left-join
  behaviour for unresolved PermIDs, multi-URL fan-out, empty output directory, and default
  output path
* `test_lock.py` — tests for `FileLock` covering local contention (two threads, timeout),
  sidecar cleanup on release, and S3 conditional-put / delete protocol (boto3-mocked)


src/idi_company_info/input.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        subset = input_df[["parent_name", "parent_cik"]].copy()
+        self.logger.info("Found %s total rows", len(subset))
+
+        # Filter out rows with null or empty values
+        subset = subset[(subset["parent_name"].notna()) & (subset["parent_cik"].astype(str) != "")]
+        self.logger.info("Found %s rows with valid CIKs", len(subset))
+
+        # Remove duplicates
+        subset["parent_cik"] = subset["parent_cik"].astype(str)  # convert identifier row to str
```
alice-duan: we zero-pad CIKs to 10 digits for the CDT input--do we want to do that here too?

src/idi_company_info/input.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        # Extract company and cik columns
+        subset = df[["company_name", "cik"]].copy()
+        self.logger.info("Found %s total rows", len(subset))
+
+        # Filter out rows with null or empty values
+        subset = subset[
+            (subset["company_name"].notna())
+            & (subset["company_name"] != "nan")
+            & (subset["cik"].astype(str) != "")
```
alice-duan: is it possible for cik to be nan? i.e. do we want a `(subset["cik"].astype(str) != "nan")` check as well here?

src/idi_company_info/input.py
alice-duan, line 1: this pattern is really neat! it sets us up well to add new inputs and is easy to track the logic for each input source 🤩 

one nonblocking thought--there is some shared logic for processing CIKs in `_extract_filter_parquet_cik`, which have pretty similar logic. i'm wondering if it makes sense to pull that shared logic out into the base class?

pulumi/Pulumi.dev.yaml
```diff
@@ -0,0 +1,30 @@
+config:
+  aws:region: us-east-2
+  idi:app_name: company-info
+  idi:bucket_name: idi-dev-ftm2j-shared-processor-storage
+  idi:output_dir: s3://idi-dev-ftm2j-shared-processor-storage/company-info/output
+  idi:input_sources:
```
alice-duan, line 6: was it intentional to exclude shareholder_tracker_cusip here?

pulumi/Pulumi.dev.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - source: shareholder_tracker_cik
+      input_file: s3://idi-dev-ftm2j-shared-processor-storage/company-info/input/shareholder_tracker_release_20251218.parquet.parquet
+      cron: cron(0 2 * * ? *)
+      batch_size: "2450"
+    - source: commercial_debt_tracker
+      input_file: s3://idi-dev-ftm2j-shared-processor-storage/database/cdt/debt-instruments/latest.parquet
+      cron: cron(30 2 * * ? *)
+      batch_size: "2450"
+    - source: corporate_subsidiaries
+      input_file: s3://idi-dev-ftm2j-shared-processor-storage/processors/corporate-structure/output/subsidiaries.parquet
```
alice-duan, line 16: i know this is where it's writing to right now, but maybe worth adding a TODO for when this changes to `database/corporate-structure/latest.parquet`, and same for the shareholder-tracker input location we wrote in [the notion data spec](https://www.notion.so/uchicago-dsi/FTM2J-Tech-Spec-36e49fa63b8980a5bc85e710e7813e6d?source=copy_link).

alternatively we can code up the correct output paths in this file now, and drop in a temp file there in the meantime so this processor can run against those locations and we don't have to come back and change these paths in code after the other processors start writing to the correct location

src/idi_company_info/input.py
```diff
@@ -0,0 +1,425 @@
...
+        if missing_columns:
+            raise ValueError(f"Required columns {missing_columns} not found in dataframe")
+
+        return input_df
+
+
+class ShareholderInputCik(Input):
+    """Input for shareholder parquet files keyed by CIK.
+
+    Expects columns: ``investor_name``, ``investor_cik``.
```
alice-duan, line 120: looking at the [data spec](https://www.notion.so/uchicago-dsi/FTM2J-Tech-Spec-36e49fa63b8980a5bc85e710e7813e6d?source=copy_link), shareholder tracker outputs `issuer_name` and `investor_cik`. we should either read from `issuer_name` here or change the spec for shareholder tracker to output `investor_name`--do you have thoughts on which would be most accurate?

alice-duan (approved):
lgtm!! the input pattern is sleek and the locking seems robust with reasonable tradeoffs

from what i understand, this is now one ECS task per input source, which runs the pipeline + aggregates `latest.parquet`. one thing i'm wondering is if we might as well have the aggregation be a separate job that is scheduled, so we have 5 tasks instead of 4, but we only run the aggregation once instead of having each task aggregate itself. pros are that we control aggregation and minimize race conditions, cons are that if it runs while another processor is still running, those changes would not get picked up until the next aggregation. not strongly opinionated on this, just wanna discuss alternatives

nikki-t:
Thanks @alice-duan - I appreciate the feedback!


> from what i understand, this is now one ECS task per input source, which runs the pipeline + aggregates latest.parquet. one thing i'm wondering is if we might as well have the aggregation be a separate job that is scheduled, so we have 5 tasks instead of 4, but we only run the aggregation once instead of having each task aggregate itself. pros are that we control aggregation and minimize race conditions, cons are that if it runs while another processor is still running, those changes would not get picked up until the next aggregation. not strongly opinionated on this, just wanna discuss alternatives

I agree but wasn't quite sure how I wanted to implement the execution of that final aggregation step. So I left the scheduled ECS task out of the Pulumi definitions. The ECS task is already defined and tested so it should be fairly trivial to create the schedule.

I think I want to hold off on implementing anything further - I would like to start to piece everything together and review the different schedules and when things execute. I think this is the only dependency so far so I am not quite sure we need an orchestration layer like AWS Step Functions quite yet but it is in the back of my mind.

src/idi_company_info/input.py
nikki-t: Sure, it has been pretty consistently zero-padded in the Subsidiary data but I will add to keep things consistent. I will add it to the Shareholder Tracker CIK operations as well.

src/idi_company_info/input.py
nikki-t: Sure, I haven't seen that in the data yet but I can add it in as a check just in case.

pulumi/Pulumi.dev.yaml
nikki-t: Yes, the shareholder tracker CUSIP operations are not finalized so I did not want to include it in the schedules for now.

pulumi/Pulumi.dev.yaml
nikki-t: I think since it is the `dev` file, it is okay to just track it as a TODO for when the output changes. I created a separate issue to track: #25

src/idi_company_info/input.py
nikki-t: Hmm, I am not sure I completely understand. There is not an `issuer_cik` so I think it has to remain `investor_cik` and `investor_name` - I think the `issuer_name` pairs with the `security_cusip`?

src/idi_company_info/input.py
nikki-t: That makes sense, and I have implemented it.

nikki-t:
Updated based on feedback and ran tests on all operations locally.

src/idi_company_info/input.py
alice-duan: got it! i meant that we should update the data spec to reflect what the shareholder tracker is outputting, since  what we documented for that currently is different from what this processor is reading from here (i.e. the spec doesn't have the field `investor_name`).

alice-duan: i added it to the shareholder tracker spec!

Issue #18, opened 2026-05-28
Query the LSEG API for business, economic and industry and ticker/exchange


nikki-t:
# context

Here is a PermID record for a company: https://permid.org/1-4296572702 with the following codes:

```
"hasPrimaryBusinessSector": "https://permid.org/1-4294952739",
 "hasPrimaryEconomicSector": "https://permid.org/1-4294952740",
 "hasPrimaryIndustryGroup": "https://permid.org/1-4294952738",
```

We may want to pull in this extra metadata as part of the company info processor. One query equals one API call so we will want to be cognizant of the daily number of calls we make.

Pull in ticker and exchange as well.

# tasks (optional)

# acceptance criteria

# out of scope (optional)

PR #24, opened 2026-06-18
Implement sector and ticker / exchange queries

nikki-t:
# Issue #18 — Sector & Ticker / Exchange Enrichment

## Summary

Extends the company-info pipeline to resolve each entity's **primary sectors** (business, economic, industry group) and its **primary quote** (ticker and exchange identifiers), and pipes the new fields through to the aggregated parquet output. PermID returns these as linked `permid.org` URLs on the entity record, so resolving them requires follow-up entity-lookup calls gated behind an `enrich_metadata` flag and budgeted against the 5K/day API quota.

- **Sectors** — each of the three sector links (`hasPrimaryBusinessSector`, `hasPrimaryEconomicSector`, `hasPrimaryIndustryGroup`) is resolved via a follow-up lookup into a label + comment.
- **Quote** — the single `hasOrganizationPrimaryQuote` link resolves to a `tr-fin:Quote` record that carries the ticker, exchange (MIC), exchange code, and RIC inline, so one call yields all of them.
- **Enrichment toggle** — `enrich_metadata` (default on) controls whether the follow-ups run; `--no-enrich-metadata` falls back to one lookup per company. Each follow-up is counted in `BatchStats.total_follow_up_calls` so the real per-run multiplier is observable.
- **Quota-driven batch sizing** — with enrichment a company costs up to 5 PermID calls (1 lookup + up to 4 follow-ups). Across the 3 daily-scheduled sources sharing one API key, `batch_size` is set to **330** (`3 × 330 × 5 = 4,950 ≤ 5,000/day`) in code defaults and both Pulumi stacks.

## Changes

| Area | Description |
|---|---|
| `src/idi_company_info/retrieval_company_info.py` | `_resolve_permid_link` follow-up helper; `_resolve_quote` → `QuoteInfo`; `_resolve_sector_name` → (label, comment); `_parse_company_info` populates the 10 new fields |
| `src/idi_company_info/types.py` | `QuoteInfo` dataclass; 10 new `CompanyResult` fields (3 sector labels, 3 sector comments, ticker, exchange, exchange_code, ric); `enrich_metadata` on `BatchConfig`/`OrchestratorConfig`; `total_follow_up_calls` stat; `batch_size` default `2450 → 330` |
| `src/idi_company_info/output.py` | `COMPANY_FIELDS` carries the new sector/quote columns into the aggregated parquet |
| `src/idi_company_info/orchestrator.py` | `--no-enrich-metadata` flag; `--batch-size` default `2450 → 330` |
| `src/idi_company_info/factory.py` | Threads `enrich_metadata` into the retrieval config |
| `src/idi_company_info/company_pipeline.py` | Logs metadata follow-up call count for the run |
| `pulumi/Pulumi.dev.yaml` / `Pulumi.prod.yaml` | `batch_size` for all 3 sources `2450 → 330` |
| `README.md` | Documents the new `--batch-size` default and quota rationale |
| `tests/` | New `test_retrieval_company_info.py` enrichment suite; updated `test_output.py` and `test_orchestrator_integration.py` for the new schema |

## Notes

- **Follow-up shape assumption** — sector/quote links are passed to the lookup as single string URLs. If the API ever returns a list or `{"@id": …}` object for these, the lookup would need to unwrap it (not handled here).
- **No `tr-org:` key fallback** — org-level keys are read bare (`hasPrimaryBusinessSector`, etc.) per the JSON-LD context; the prefixed form is not used as a fallback.
- **Re-tuning** — `330` assumes the worst case (every company hits all 4 follow-ups). Once `total_follow_up_calls` from production shows the real multiplier (usually lower), `batch_size` can be raised to drain backfill faster.

## Test Plan

- [X] `uv run pytest` passes (enrichment, output, and orchestrator-integration suites)
- [x] With enrichment on, a sample entity resolves non-null sector labels and ticker/exchange/exchange_code/ric
- [x] `--no-enrich-metadata` makes exactly one lookup per company and leaves the 10 new fields null
- [x] Aggregated `latest.parquet` contains the new sector and quote columns
- [x] A full `batch_size: 330` run logs `total_follow_up_calls` and stays within the 5K/day PermID quota across all 3 sources


src/idi_company_info/types.py
```diff
@@ -128,9 +153,10 @@ class OrchestratorConfig:
     input_type: InputSource
     api_key: str
     geonames_user: str
-    batch_size: int = 2450
+    batch_size: int = 330
```
alice-duan: suggestion: instead of hardcoding this calculated batch size based on the assumption of 5 calls per company, what if we budget by actual call counts? i.e. batch size conceptually becomes something more like `max_requests`. we would need to count the number of ~follow up~ requests made, and some things like making sure a batch doesn't cut off midway through a company's up-to-5 calls. but we would get more bang for our buck with to 5k limit

src/idi_company_info/retrieval_company_info.py
```diff
@@ -172,21 +175,46 @@ def _retrieve_entity_company_info(
...
         Returns:
             A flat company-info entry (API fields + last_processed). The permid_url is
             the dict key — there is no ``search``/``result`` envelope.
         """
+        quote_info = self._resolve_quote(response.get("hasOrganizationPrimaryQuote"), batch_stats)
+
+        bus_label, bus_comment = self._resolve_sector_name(
+            response.get("hasPrimaryBusinessSector"), batch_stats
+        )
+        eco_label, eco_comment = self._resolve_sector_name(
```
alice-duan, line 264: i googled this briefly and it seems there is a pretty small number of possible sectors and groups (<100). that would mean a lot of companies have the same value for these fields, so we could benefit from memoizing these results in a dict, which stores the value returned for each permid we resolve at this stage: `{permid_link: value}`. and for call to `_resolve_...`, we can check this cache before making a request to LSEG. so, compounded with the suggestion to batch by # requests instead of # companies, we could process a lot more per batch. 

we could either 
1. store the memoized cache in memory per-run, or 
2. flush to s3 at the end of every run and load it back in at the beginning of the next run--then we could almost always bypass the 3 API calls for sector and group

if the number of sectors is really small (<100), then it wouldn't really matter which approach we take, because in each batch we max out at n=(# of unique sectors) requests for this part of the enrichment process

src/idi_company_info/types.py
nikki-t: I have implemented this in the codebase.

src/idi_company_info/retrieval_company_info.py
nikki-t: I also implemented this into the codebase. I decided to flush to S3 since we may find the sector categories useful in some other context and this allows us to run on smaller batches, cache the results, and pull them back in. I am going to deploy and run in the cloud as final test.

nikki-t:
Confirmed recent changes deployed and executed in the cloud successfully.

